PR #2, opened 2017-12-14
EID-257: Remove blanket catch of exception

adityapahuja:
SAML engine was updated to use correct name id policy (persistent instead of unspecified) for eIDAS Authn Request. eIDAS Authn Request and Verify Authn Request now have same name id format. As a result of this change, eIDAS Authn Response now contains correct name id format (persistent). It will pass name id format validation without errors. This means that the blanket catch of exception in policy is no longer needed.

Authors: @adityapahuja

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/services/AuthnResponseFromCountryService.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/services/AuthnResponseFromCountryService.java
@@ -69,18 +69,13 @@ public class AuthnResponseFromCountryService {
         String matchingServiceEntityId = stateController.getMatchingServiceEntityId();
         stateController.validateCountryIsIn(countriesService.getCountries(sessionId));
 
-        // TODO: Bug in CEF Reference 1.1. Remove this try..catch when EID-177 CEF Reference 1.3 is deployed.
-        try {
-            SamlAuthnResponseTranslatorDto responseToTranslate = samlAuthnResponseTranslatorDtoFactory.fromSamlAuthnResponseContainerDto(responseFromCountry, matchingServiceEntityId);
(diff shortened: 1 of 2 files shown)


hub/policy/src/main/java/uk/gov/ida/hub/policy/services/AuthnResponseFromCountryService.java
```diff
@@ -69,18 +69,13 @@ public ResponseAction receiveAuthnResponseFromCountry(SessionId sessionId,
...
-            InboundResponseFromCountry translatedResponse = samlEngineProxy.translateAuthnResponseFromCountry(responseToTranslate);
-            validateTranslatedResponse(stateController, translatedResponse);
-            EidasAttributeQueryRequestDto eidasAttributeQueryRequestDto = getEidasAttributeQueryRequestDto(stateController, translatedResponse);
-            stateController.transitionToEidasCycle0And1MatchRequestSentState(eidasAttributeQueryRequestDto, responseFromCountry.getPrincipalIPAddressAsSeenByHub(), translatedResponse.getIssuer());
-            AttributeQueryContainerDto aqr = samlEngineProxy.generateEidasAttributeQuery(eidasAttributeQueryRequestDto);
-            samlSoapProxyProxy.sendHubMatchingServiceRequest(sessionId, getAttributeQueryRequest(aqr));
-        } catch (RuntimeException e) {
-            LOG.info("Exception when translating country response", e);
-        }
+        SamlAuthnResponseTranslatorDto responseToTranslate = samlAuthnResponseTranslatorDtoFactory.fromSamlAuthnResponseContainerDto(responseFromCountry, matchingServiceEntityId);
```
IreneLau-GDS: This makes me soooooooooo happy! Without the try-catch!

PR #4, opened 2017-12-22
TT-1603: Allow hub-config to return IDPs relevant to a particular RP …

CharlesIC:
…and LoA.

Implemented logic in hub to filter available IDPs on both transaction entity and supported LOA
IdentityProviderConfigEntityData: renamed isOnboarding(loa) to isOnboardingAtLoa(loa); added unit tests for the class

Authors: @CharlesIC, @hugh-emerson

### DIFF
--- a/hub/config/src/main/java/uk/gov/ida/hub/config/domain/IdentityProviderConfigEntityData.java
+++ b/hub/config/src/main/java/uk/gov/ida/hub/config/domain/IdentityProviderConfigEntityData.java
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
 import org.apache.commons.collections.ListUtils;
-import org.apache.commons.io.FileUtils;
 import uk.gov.ida.hub.config.ConfigEntityData;
 
 import javax.validation.Valid;
@@ -66,6 +65,11 @@ public class IdentityProviderConfigEntityData implements ConfigEntityData {
         return entityId;
     }
 
+    @Override
+    public String toString() {
+        return String.format("%s | %s", this.getEntityId(), this.getSimpleId());
+    }
+
     public List<LevelOfAssurance> getSupportedLevelsOfAssurance() {
         return supportedLevelsOfAssurance;
     }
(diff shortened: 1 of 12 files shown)


hub/config/src/main/java/uk/gov/ida/hub/config/domain/IdentityProviderConfigEntityData.java
```diff
@@ -66,6 +65,11 @@ public String getEntityId() {
         return entityId;
     }
 
+    @Override
+    public String toString() {
+        return String.format("%s | %s", this.getEntityId(), this.getSimpleId());
```
hugh-emerson: Why did you add this toString method?

hub/config/src/test/java/uk/gov/ida/hub/config/domain/filters/IdpPredicateFactoryPredicatesTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public void shouldReturnIdpsForLoaForNonOnboardingTransactionEntity() {
+        final Set<Predicate<IdentityProviderConfigEntityData>> loa1Predicate = idpPredicateFactory
+                .createPredicatesForTransactionEntityAndLoa(transactionEntityNonOnboarding, LevelOfAssurance.LEVEL_1);
+
+        final Set<IdentityProviderConfigEntityData> filteredIdps = getFilteredIdps(allIdps, loa1Predicate);
+
+        final IdentityProviderConfigEntityData[] expectedFilteredIdps = {nonOnboardingLoa1Idp, nonOnboardingAllLevelsIdp,
+                onboardingLoa2Idp, onboardingLoa2IdpOtherOnboardingEntity};
+
+        assertThat(filteredIdps).hasSameSizeAs(expectedFilteredIdps);
```
hugh-emerson: hasSameSizeAs should be unnecessary when you are asserting containsOnly.

From docs for containsOnly:
`Verifies that the actual group contains only the given values and nothing else`

hub/config/src/main/java/uk/gov/ida/hub/config/domain/IdentityProviderConfigEntityData.java
CharlesIC: Yes, it was for the unit tests as otherwise the entities couldn't be identified in the error messages. Also helpful if we ever get an exception so the useful info would show up in the message instead of the random numeric id generated by JVM

hub/config/src/test/java/uk/gov/ida/hub/config/domain/filters/IdpPredicateFactoryPredicatesTest.java
CharlesIC: Just me being extra pedantic.... The extra check can be removed if it's confusing but I don't think it hurts to have it there.

PR #9, opened 2018-01-12
TT-1613: Switch hub-policy to fetch valid IDPs using LoA in addition to Transaction.

CharlesIC:


hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/AuthnFailedErrorStateController.java
```diff
@@ -1,12 +1,7 @@
 package uk.gov.ida.hub.policy.domain.controller;
 
 import com.google.common.base.Optional;
-import uk.gov.ida.hub.policy.domain.AuthnRequestSignInProcess;
-import uk.gov.ida.hub.policy.domain.FailureResponseDetails;
-import uk.gov.ida.hub.policy.domain.ResponseFromHub;
-import uk.gov.ida.hub.policy.domain.ResponseFromHubFactory;
-import uk.gov.ida.hub.policy.domain.StateController;
-import uk.gov.ida.hub.policy.domain.StateTransitionAction;
+import uk.gov.ida.hub.policy.domain.*;
```
MichaelWalker: As a tiny style point - the convention here is to prefer explicit imports to wildcard imports, so we should probably undo this.

(I think it makes sense too, especially as many of our libraries have clashing namespaces)

willp-bl: Some help for how to set it up here - https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

CharlesIC: It's IntelliJ's default behaviour to "optimise" imports by doing this - and I've not changed that. Thanks @willp-bl for the link.

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/IdpSelected.java
```diff
@@ -23,10 +25,11 @@ public IdpSelected(String selectedIdpEntityId, String principalIpAddress) {
         this.principalIpAddress = principalIpAddress;
     }
 
-    public IdpSelected(String selectedIdpEntityId, String principalIpAddress, Boolean registration) {
+    public IdpSelected(String selectedIdpEntityId, String principalIpAddress, Boolean registration, LevelOfAssurance requestedLoa) {
```
MichaelWalker: Hmm - this constructor is starting to take quite a few params.

I tend to find that calling long constructors in Java gets very confusing very fast (no named params 😭 - although to be fair Intellij is trying to work around this now).
The standard Java pattern is to use a builder to wrap the constructor, ie

```
new Foo(true, true, false, true);
``` 
becomes
```
aFoo()
  .withEnabled(true)
  .withImage(true)
  .withRegistration(false)
  .withParent(true)
  .build()
```
which is quite a lot more obvious (if a bit more verbose).

I don't think that this is a pattern that we've used anywhere in verify though (we only seem to use builders in our test code) and I'm not sure that this specific case is a good enough reason to add a whole new pattern.

What do you think? If you have another idea that would also be great!
(And sorry if you know all this already - let me know and I'll shut up a bit! 😄 )

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/IdpSelector.java
```diff
@@ -18,6 +18,7 @@ private IdpSelector() {}
     public static IdpSelectedState buildIdpSelectedState(IdpSelectingState state,
```
MichaelWalker: As with the class, I'm not a huge fan of the long signatures.
If you have any ideas that would be awesome.

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
```diff
@@ -55,6 +51,7 @@
 public class AuthnRequestFromTransactionResourceIntegrationTest {
     private static String TEST_SESSION_RESOURCE_PATH = Urls.PolicyUrls.POLICY_ROOT + "test";
     private static final Boolean REGISTERING = TRUE;
+    private static final LevelOfAssurance REQUESTED_LOA = LevelOfAssurance.LEVEL_2;
```
MichaelWalker: I'm not sure that this adds very much. `LevelOfAssurance.LEVEL_2` is already a constant, and the tests would all be a little clearer if we used it directly (rather than hiding it in a class constant)
(The same goes for boolean above to be honest)

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/MatchingServiceResourcesIntegrationTest.java
```diff
@@ -101,6 +101,7 @@
     private SamlResponseWithAuthnRequestInformationDto translatedAuthnRequest;
     private SamlAuthnRequestContainerDto rpSamlRequest;
     private boolean registering = true;
+    private LevelOfAssurance requestedLoa = LevelOfAssurance.LEVEL_2;
```
MichaelWalker: this (and the line above) could be `static final`, but as above, perhaps it doesn't need to exist at all?

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
MichaelWalker: You could also statically import `LEVEL_2`, which would allow you to need to bother writing `LevelOfAssurance.` throughout these classes

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/AuthnRequestSignInProcess.java
```diff
@@ -13,19 +10,13 @@ private AuthnRequestSignInProcess() {
     }
 
     public AuthnRequestSignInProcess(
```
MichaelWalker: nice spot that we can kill all of this stuff!

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/IdpSelected.java
CharlesIC: Fair point but I think it's fine in this case because 1) It's got _only_ 4 parameters (doesn't seem to be too many to me) 2) This constructor is only used in code in tests, and it's really there only for the JSON/Javax validation libraries when translating POST messages from frontend into Java objects

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/IdpSelector.java
CharlesIC: Normally you'd have some fields initialised when instantiating an object but this is a static method so it needs to have all of its parameters passed in. Unless it's worth/possible to inject the two proxy objects instead?

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/rest/Cycle3DTO.java
```diff
@@ -91,4 +92,8 @@ public String getEncryptedMatchingDatasetAssertion() {
     public boolean getTransactionSupportsEidas() {
         return transactionSupportsEidas;
     }
+
+    public boolean isRegistering() {
```
MichaelWalker: It looks like this is always going to be false.

a) is that right?
b) if so, can we just `return false;` and skip the private field?

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
CharlesIC: I've seen this done in many integration tests - I think it's because of consistency? So the same value is used across the setup and in the test cases. Perhaps a static import would be more succinct.

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/rest/Cycle3DTO.java
CharlesIC: Fair point. And the same goes for transactionSupportsEidas and getTransactionSupportsEidas just above

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/support/ConfigStubRule.java
```diff
@@ -68,7 +72,7 @@ public void setUpStubForLevelsOfAssurance(String entityId) throws JsonProcessing
         String uri = UriBuilder.fromPath(Urls.ConfigUrls.LEVELS_OF_ASSURANCE_RESOURCE)
             .build(StringEncoding.urlEncode(entityId).replace("+", "%20"))
             .getPath();
-        register(uri, OK, asList(LevelOfAssurance.LEVEL_2, LevelOfAssurance.LEVEL_2));
+        register(uri, OK, asList(LevelOfAssurance.LEVEL_1, LevelOfAssurance.LEVEL_2));
```
MichaelWalker: good spot! 😆

hub/policy/src/test/java/uk/gov/ida/hub/policy/domain/controller/IdpSelectedStateControllerTest.java
```diff
@@ -67,7 +47,7 @@
     private static final SessionId NEW_SESSION_ID = aSessionId().build();
     private static final List<LevelOfAssurance> levelsOfAssurance = asList(LevelOfAssurance.LEVEL_1, LevelOfAssurance.LEVEL_2);
     private static final LevelOfAssurance providedLevelOfAssurance = LevelOfAssurance.LEVEL_2;
-    private static final Optional<String> principalIpAddressAsSeenByIdp = Optional.fromNullable("principal-ip-address-from-idp");
+    private static final Optional<String> principalIpAddressAsSeenByIdp = Optional.of("principal-ip-address-from-idp");
```
MichaelWalker: Convention is for static final things to be UPPER_CASE.
(we seem to not be doing that in some of this class though - would you mind fixing it?)

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/IdpSelector.java
MichaelWalker: Named parameters in other languages make this kind of thing much nicer.
You could use the builder pattern, but then you have to find a way of telling / ensuring that all the parameters have been passed in.

I think just leave it for now - I don't have a solution which is any less awkward that what it already was!

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/rest/Cycle3DTO.java
CharlesIC: Actually, these cannot be removed as this is a DTO object used in integration tests and these fields are expected to be present by the JSON validator. Adding them back in.

PR #11, opened 2018-01-12
TT-1613: Switch hub-policy to fetch valid IDPs using LoA in addition …

CharlesIC:
…to Transaction

Add a nullable requestedLoa parameter to IdpSelected class to prepare for the corresponding frontend release (TT-1613 release 1/3)

Authors: @CharlesIC, @MichaelWalker

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/IdpSelected.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/IdpSelected.java
@@ -12,6 +12,8 @@ public class IdpSelected {
     private String principalIpAddress;
     @NotNull
     private Boolean registration;
+    // TODO: Make NotNull after the corresponding hub release
+    private LevelOfAssurance requestedLoa;
 
     // Required for JAXB
     @SuppressWarnings("unused")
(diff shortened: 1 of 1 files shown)


hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/IdpSelected.java
```diff
@@ -23,10 +25,19 @@ public IdpSelected(String selectedIdpEntityId, String principalIpAddress) {
         this.principalIpAddress = principalIpAddress;
     }
 
+    // TODO: Remove this construtor after the corresponding hub release
     public IdpSelected(String selectedIdpEntityId, String principalIpAddress, Boolean registration) {
```
MichaelWalker: not worth changing (as the code is about to be deleted), but for the future, a slightly better pattern for this is

```
public Foo(Thing a, Thing b) {
    return new Foo(a, b, null);
}

public Foo(Thing a, Thing b, Thing c) {
    this.a = a;
    this.b = b;
    this.c = c;
}
```

That way we only have to do the setting once, we cut down the code duplication, and the intention is a little clearer.

PR #23, opened 2018-02-12
TT-1613: Use new config endpoint

CharlesIC:
Use the new parameters (isRegistration and requestedLoa) in the modified state classes to call the new config endpoint. At this point the app will switch to storing the new definitions in the cache, and the transitional classes will be removed in the next release.


hub/config/src/main/java/uk/gov/ida/hub/config/Urls.java
```diff
@@ -36,7 +36,8 @@
         String ENABLED_IDENTITY_PROVIDERS_PATH = "/enabled";
         String ENABLED_IDENTITY_PROVIDERS_RESOURCE = IDENTITY_PROVIDER_ROOT + ENABLED_IDENTITY_PROVIDERS_PATH;
         String ENABLED_IDENTITY_PROVIDERS_PARAM_PATH = ENTITY_ID_PATH_PARAM + "/enabled"; //deprecated
-        String ENABLED_ID_PROVIDERS_FOR_LOA_PATH = ENTITY_ID_PATH_PARAM + LEVEL_OF_ASSURANCE_PATH_PARAM + "/enabled";
+        String ENABLED_IDENTITY_PROVIDERS_PARAM_PATH_RESOURCE = IDENTITY_PROVIDER_ROOT + ENABLED_IDENTITY_PROVIDERS_PARAM_PATH; // deprecated
```
MichaelWalker: I think you can `@Deprecated` a field if that sounds fun?

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
```diff
@@ -182,21 +180,11 @@ public void getSignInProcessDto_shouldReturnSignInDetailsDto(){
...
-        );
-
-        List<String> expectedEntityIds = IdpSelectedStateBuilder.anIdpSelectedState().build().getAvailableIdentityProviderEntityIds();
-
         Response response = getAuthRequestSignInProcess(session);
 
         assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
         AuthnRequestSignInDetailsDto entity = response.readEntity(AuthnRequestSignInDetailsDto.class);
         assertThat(entity.getRequestIssuerId()).isEqualTo(samlResponse.getIssuer());
-        assertThat(entity.getAvailableIdentityProviderEntityIds()).isEqualTo(expectedEntityIds);
```
MichaelWalker: how come we can get rid of these assertions

CharlesIC: Becasue the AuthnRequestSignInDetailsDto object no longer has the methods/fields getAvailableIdentityProviderEntityIds() and getAvailableIdentityProviders(). They were not really used, and this information is now retrieved from the new config endpoint instead (using the proxy).

hub/config/src/main/java/uk/gov/ida/hub/config/Urls.java
CharlesIC: You can! How fun. Added to the review commit

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
MichaelWalker: perfect. Thanks

willp-bl: It would be good to talk more about the difference between storing the IDPs in the session state vs dynamically getting them from the config service (and the relative merits)

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/EidasMatchRequestSentStateController.java
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.ida.hub.policy.domain.StateTransitionAction;
+import uk.gov.ida.hub.policy.domain.state.EidasMatchRequestSentState;
+import uk.gov.ida.hub.policy.domain.state.EidasSuccessfulMatchState;
+import uk.gov.ida.hub.policy.logging.EventSinkHubEventLogger;
+import uk.gov.ida.hub.policy.services.AttributeQueryService;
+import uk.gov.ida.hub.policy.validators.LevelOfAssuranceValidator;
+
+public abstract class EidasMatchRequestSentStateController<T extends EidasMatchRequestSentState> extends AbstractMatchRequestSentStateController<T, EidasSuccessfulMatchState> {
+
+    public EidasMatchRequestSentStateController(
```
MichaelWalker: Does this constructor do anything? It looks like it just just passes things straight to the parent.
Isn't that the same thing as what would happen if we just didn't write it?

CharlesIC: You're right, but this a question for the eIDAS team. Since this is work in progress, they might add more properties to the eIDAS states and logic to their controllers. I know they (i.e. Hugh) did it to keep it separate from the non-eidas stuff (for now?)

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
CharlesIC: In general, we store these properties in the states themselves whenever we can. For instance, we can only get a list of available IDPs once we get to the IdpSelected stated, but not before, we don't have the required information to obtain that list. In this specific case, AuthnRequestSignInDetailsDto had this information stored but it was only used in an integration test and not anywhere in the code, so I removed it.

willp-bl: As I understood it the available IDPs were cached in the session state when the session was created and Policy used that store directly afterwards (instead of Config), as well as storing the IDP that was selected.  Things must have changed since I last looked. 

FWIW saml-engine, however, uses IDP info directly from Config.

PR #26, opened 2018-02-13
ABD-35: Update to log all events using event emitter

adityapahuja:
Update policy, saml proxy and saml soap proxy to log all other events (e.g. error events and external communications events) using event emitter library.

Authors: @adityapahuja @MichaelWalker


build.gradle
```diff
@@ -21,6 +21,7 @@ ext {
 
 def dependencyVersions = [
             ida_utils:'323',
+            rest_utils:'330',
```
willp-bl: any reason to use rest-utils at the same time as ida-utils? it's the same lib...

willp-bl: we shouldn't let the versions drift if they are both required for some reason

willp-bl: (I've not reviewed this PR)

MichaelWalker: Agree that we shouldn't let the versions drift.

In this case upgrading security utils to 330 breaks all kinds of things in some quite non-trivial ways. Andy Paine has been working on getting a fix for that, but for now we are keen to commit this code in a non breaking way, (as its a large enough merge as it is and we'd like to avoid long lived branches)

These should get back in sync again soon - just wanted to be able to commit this work before it causes conflicts.

willp-bl: Could you add a comment to that effect so people know it's temporary?

adityapahuja: This has been done. Please see https://github.com/alphagov/verify-hub/commit/d232e6a84ce73effb46e820b89be782b4a8dac4f.

hub/policy/src/main/java/uk/gov/ida/hub/policy/exception/IdpDisabledExceptionMapper.java
```diff
@@ -1,52 +1,41 @@
...
 
-import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.ida.common.ErrorStatusDto;
 import uk.gov.ida.common.ExceptionType;
 import uk.gov.ida.common.ServiceInfoConfiguration;
 import uk.gov.ida.eventsink.EventDetailsKey;
 import uk.gov.ida.eventsink.EventSinkHubEventConstants;
+import uk.gov.ida.eventsink.EventSinkProxy;
```
phantas: Why are we leaving eventsink here too?

hub/policy/src/main/java/uk/gov/ida/hub/policy/exception/PolicyApplicationExceptionMapper.java
```diff
@@ -61,18 +58,7 @@ private boolean auditException(final ApplicationException exception) {
...
-
-            if (getSessionId().isPresent()) {
-                eventSinkMessageSenderFacade.audit(exception, exception.getErrorId(), getSessionId().get(), eventDetails);
-            } else {
-                eventSinkMessageSenderFacade.audit(exception, exception.getErrorId(), NO_SESSION_CONTEXT_IN_ERROR,
-                        eventDetails
-                );
-            }
-
+            eventLogger.logErrorEvent(exception.getErrorId(), getSessionId().or(NO_SESSION_CONTEXT_IN_ERROR), exception.getMessage(), exception.getUri().or(URI.create("uri-not-present")).toASCIIString());
```
phantas: ❤️ simplification

phantas: but I would break this line in multilines, it's too long and hard to see where the "or"s relate to

adityapahuja: I have fixed it in a new pull request (https://github.com/alphagov/verify-hub/pull/30).

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
```diff
@@ -9,6 +9,8 @@
 import uk.gov.ida.common.ServiceInfoConfiguration;
 import uk.gov.ida.common.shared.security.IdGenerator;
 import uk.gov.ida.eventsink.EventSink;
+import uk.gov.ida.eventsink.EventSinkHttpProxy;
+import uk.gov.ida.eventsink.EventSinkProxy;
```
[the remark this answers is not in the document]
adityapahuja: Event Sink is still required since policy needs to send events to Event Sink for audit and billing purposes.

PR #27, opened 2018-02-13
ABD-35: Add a feature toggle to send events to a recording system

adityapahuja:
Adding a configurable toggle to policy, saml soap proxy and saml proxy to send or not to send events to a recording system. The default is not to send events to the recording system.

Authors: @adityapahuja

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/logging/HubEventLogger.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/logging/HubEventLogger.java
@@ -10,6 +10,7 @@ import uk.gov.ida.eventemitter.EventEmitter;
 import uk.gov.ida.eventsink.EventDetailsKey;
 import uk.gov.ida.eventsink.EventSinkHubEventConstants;
 import uk.gov.ida.eventsink.EventSinkProxy;
+import uk.gov.ida.hub.policy.PolicyConfiguration;
 import uk.gov.ida.hub.policy.contracts.SamlResponseWithAuthnRequestInformationDto;
 import uk.gov.ida.hub.policy.domain.EventSinkHubEvent;
 import uk.gov.ida.hub.policy.domain.FraudDetectedDetails;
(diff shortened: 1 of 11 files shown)


hub/policy/src/main/java/uk/gov/ida/hub/policy/logging/HubEventLogger.java
```diff
@@ -199,7 +205,10 @@ public void logRequestFromHub(SessionId sessionId, String transactionEntityId) {
                 details);
 
         eventSinkProxy.logHubEvent(eventSinkHubEvent);
-        eventEmitter.record(eventSinkHubEvent);
+
+        if (sendToRecordingSystem) {
```
MichaelWalker: Why not have the logic about 'to send or not to send' live in the EventEmitter? That way we reduce the code duplication?

willp-bl:
I've not read the PR but we must continue to send to event sink, is it the new recording system that is off by default?

hub/policy/src/main/java/uk/gov/ida/hub/policy/logging/HubEventLogger.java
```diff
@@ -75,12 +76,17 @@
...
     private final ServiceInfoConfiguration serviceInfo;
     private final EventEmitter eventEmitter;
+    private final boolean sendToRecordingSystem;
 
     @Inject
-    public HubEventLogger(ServiceInfoConfiguration serviceInfo, EventSinkProxy eventSinkProxy, EventEmitter eventEmitter) {
+    public HubEventLogger(ServiceInfoConfiguration serviceInfo,
+                          EventSinkProxy eventSinkProxy,
+                          EventEmitter eventEmitter,
+                          PolicyConfiguration configuration) {
```
MichaelWalker: Pulling in the whole configuration just for a single boolean feels slightly overkill. Is there a nice way to only pass in the things that we care about?

hub/policy/src/main/java/uk/gov/ida/hub/policy/logging/HubEventLogger.java
adityapahuja: Good idea, I will make the changes to Event Emitter.

adityapahuja:
@willp-bl This change does not affect Hub sending events to Event Sink. In other words, when the toggle is set to not to send events to a new recording system, Hub will still continue to send events to Event Sink. The new recording system is not set up in production yet but is available in test environment. That's why we added the toggle. I have created a new pull request. Please see https://github.com/alphagov/verify-hub/pull/30.

PR #28, opened 2018-02-13
BAU: Upgrade security utils

andy-paine:
Due to changes to the way DeserializablePublicKeyConfiguration works,
fed-config DTOs can no longer use that class to get certificates.
Due to an outstanding bug, we cannot use the specific subtype so
this commit ports the relevant class (and merges the abstract super)
to perform the same behaviour locally in hub.
To allow the DTO getters to remain the same, certificates are new-ed
inside the getters. This requires the base Certificate class to be
comparable to prevent referencing issues so an equals and hashCode
method were added

There is a plan to eventually migrate these DTOs to a database rather
than file backed system in which case this issue will be redundant

### DIFF
--- a/hub/config/src/main/java/uk/gov/ida/hub/config/domain/CertificateDetails.java
+++ b/hub/config/src/main/java/uk/gov/ida/hub/config/domain/CertificateDetails.java
@@ -24,7 +24,7 @@ public class CertificateDetails {
         this.isEnabled = isEnabled;
     }
 
-    public static CertificateDetails aCertifcateDetail(String issuerId, Certificate certificate, FederationEntityType federationEntityType) {
+    public static CertificateDetails aCertificateDetail(String issuerId, Certificate certificate, FederationEntityType federationEntityType) {
         return new CertificateDetails(issuerId, certificate, federationEntityType);
     }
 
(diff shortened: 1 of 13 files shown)


hub/config/src/main/java/uk/gov/ida/hub/config/domain/CertificateDetails.java
```diff
@@ -24,7 +24,7 @@ public CertificateDetails(String issuerId, Certificate certificate, FederationEn
         this.isEnabled = isEnabled;
     }
 
-    public static CertificateDetails aCertifcateDetail(String issuerId, Certificate certificate, FederationEntityType federationEntityType) {
+    public static CertificateDetails aCertificateDetail(String issuerId, Certificate certificate, FederationEntityType federationEntityType) {
```
andy-paine: Typo fixed

hub/config/src/main/java/uk/gov/ida/hub/config/domain/X509Certificate.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static java.text.MessageFormat.format;
+import static uk.gov.ida.common.shared.security.Certificate.BEGIN_CERT;
+import static uk.gov.ida.common.shared.security.Certificate.END_CERT;
+
+/* This class is modified from X509CertificateConfiguration in security-utils due to an
+* outstanding bug https://github.com/FasterXML/jackson-databind/issues/1358 which means
+* we cannot directly deserialize to X509CertificateConfiguration */
+public class X509Certificate {
+    @JsonProperty
```
andy-paine: The AppRule base class serializes the DTOs that use this so it needs some JsonProperty annotations

PR #33, opened 2018-02-26
ABD-35: Upgrade to use latest version of rest-utils

adityapahuja:
The latest version of rest-utils supports the ability to send encrypted events to either a source queue or a stub queue. If the source queue name is not present, encrypted event messages will be sent to the stub queue. If the source queue is present, encrypted event messages will be sent to the source queue. The stub queue will just write the encrypted event message to the standard output.

Authors: @adityapahuja

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,10 +21,10 @@ ext {
 
 def dependencyVersions = [
             ida_utils:'323',
-            // rest_utils:'330' has been added to dependency temporarily for enabling Hub to send events using event emitter.
-            // Upgrading ida_utils to 330 breaks verify-hub. The fix for this issue is still being developed. Once the fix is completed,
-            // rest_utils will be removed from the dependency.
-            rest_utils:'330',
+            // rest_utils has been added to dependency temporarily for enabling Hub to send events using event emitter.
+            // Upgrading ida_utils to the latest version breaks verify-hub. The fix for this issue is still being developed.
+            // Once the fix is completed, rest_utils will be removed from the dependency.
+            rest_utils:'332',
             dropwizard:'1.1.4',
             dropwizard_infinispan:'1.1.4-41',
             pact:'3.5.6',
@@ -99,7 +99,7 @@ subprojects {
 
         rest_utils "uk.gov.ida:rest-utils:2.0.0-$dependencyVersions.rest_utils"
 
-        verify_event_emitter "uk.gov.ida:verify-event-emitter:0.0.1-10"
+        verify_event_emitter "uk.gov.ida:verify-event-emitter:0.0.1-30"
 
         ida_utils "uk.gov.ida:common-utils:2.0.0-$dependencyVersions.ida_utils",
                 "uk.gov.ida:security-utils:2.0.0-$dependencyVersions.ida_utils"
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
@@ -8,7 +8,7 @@ import io.dropwizard.setup.Environment;
 import org.joda.time.DateTime;
 import uk.gov.ida.common.ServiceInfoConfiguration;
 import uk.gov.ida.common.shared.security.IdGenerator;
-import uk.gov.ida.eventemitter.EventEmitter;
+import uk.gov.ida.eventemitter.Configuration;
 import uk.gov.ida.eventsink.EventSink;
 import uk.gov.ida.eventsink.EventSinkHttpProxy;
 import uk.gov.ida.eventsink.EventSinkProxy;
@@ -53,6 +53,7 @@ import javax.inject.Singleton;
 import javax.ws.rs.client.Client;
 import java.net.URI;
 import java.security.KeyStore;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentMap;
 
 public class PolicyModule extends AbstractModule {
@@ -92,9 +93,8 @@ public class PolicyModule extends AbstractModule {
     }
 
     @Provides
-    @Singleton
-    public EventEmitter getEventEmitter(PolicyConfiguration configuration) {
-        return new EventEmitter(configuration.getSendToRecordingSystem());
+    private Optional<Configuration> getEventEmitterConfiguration(final PolicyConfiguration configuration) {
+        return Optional.ofNullable(configuration.getEventEmitterConfiguration());
     }
 
     @Provides
--- a/hub/policy/src/test/java/uk/gov/ida/hub/policy/PolicyApplicationTest.java
+++ b/hub/policy/src/test/java/uk/gov/ida/hub/policy/PolicyApplicationTest.java
@@ -25,9 +25,9 @@ public class PolicyApplicationTest {
     }
 
     @Test
-    public void shouldReturnSendToRecordingSystemAsFalseByDefault() throws Exception {
+    public void shouldReturnEventEmitterConfigurationAsNullByDefault() throws Exception {
         application.run(config, environment);
 
-        assertThat(config.getSendToRecordingSystem()).isFalse();
+        assertThat(config.getEventEmitterConfiguration()).isNull();
     }
 }
--- a/hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/SessionTimeoutIntegrationTests.java
+++ b/hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/SessionTimeoutIntegrationTests.java
@@ -65,8 +65,7 @@ public class SessionTimeoutIntegrationTests {
             ConfigOverride.config("samlEngineUri", samlEngineStub.baseUri().build().toASCIIString()),
             ConfigOverride.config("configUri", configStub.baseUri().build().toASCIIString()),
             ConfigOverride.config("eventSinkUri", eventSinkStub.baseUri().build().toASCIIString()),
-            ConfigOverride.config("timeoutPeriod", format("{0}m", SOME_TIMEOUT)),
-            ConfigOverride.config("sendToRecordingSystem", "true"));
+            ConfigOverride.config("timeoutPeriod", format("{0}m", SOME_TIMEOUT)));
 
     private SamlAuthnRequestContainerDto samlRequest;
 
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/EventEmitterConfiguration.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/EventEmitterConfiguration.java
@@ -1,0 +1,24 @@
+package uk.gov.ida.hub.policy;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.ida.eventemitter.Configuration;
+
+import javax.validation.Valid;
+
+public class EventEmitterConfiguration implements Configuration {
+
+    @Valid
+    @JsonProperty
+    private String sourceQueueName;
+
+    private EventEmitterConfiguration() { }
+
+    public EventEmitterConfiguration(String sourceQueueName) {
+        this.sourceQueueName = sourceQueueName;
+    }
+
+    @Override
+    public String getSourceQueueName() {
+        return sourceQueueName;
+    }
+}
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyApplication.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyApplication.java
@@ -12,6 +12,7 @@ import io.dropwizard.setup.Environment;
 import uk.gov.ida.bundles.LoggingBundle;
 import uk.gov.ida.bundles.MonitoringBundle;
 import uk.gov.ida.bundles.ServiceStatusBundle;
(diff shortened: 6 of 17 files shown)


build.gradle
```diff
@@ -24,7 +24,7 @@ def dependencyVersions = [
             // rest_utils:'330' has been added to dependency temporarily for enabling Hub to send events using event emitter.
```
MichaelWalker: can we get rid of this comment yet, or does it just need to be updated to have the right version referenced?

build.gradle
```diff
@@ -24,7 +24,7 @@ def dependencyVersions = [
...
             // Upgrading ida_utils to 330 breaks verify-hub. The fix for this issue is still being developed. Once the fix is completed,
             // rest_utils will be removed from the dependency.
-            rest_utils:'330',
+            rest_utils:'332',
```
MichaelWalker: Do we also depend on `event-emitter` directly (I can't see it immediately).
If we don't then we definitely should!

MichaelWalker: ahh - I tell a lie - I can see we delete it below...
Why are we dropping the dependency on event-emitter?

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
```diff
@@ -92,9 +93,8 @@ protected void configure() {
     }
 
     @Provides
-    @Singleton
-    public EventEmitter getEventEmitter(PolicyConfiguration configuration) {
-        return new EventEmitter(configuration.getSendToRecordingSystem());
+    private Optional<Configuration> getEventEmitterConfiguration(final PolicyConfiguration configuration) {
+        return Optional.ofNullable(configuration.getEventEmitterConfiguration());
```
MichaelWalker: Why does this need to be an optional?

hub/policy/src/test/java/uk/gov/ida/hub/policy/PolicyApplicationTest.java
```diff
@@ -25,9 +25,9 @@ public void setUp() throws Exception {
     }
 
     @Test
-    public void shouldReturnSendToRecordingSystemAsFalseByDefault() throws Exception {
+    public void shouldReturnEventEmitterConfigurationAsNullByDefault() throws Exception {
```
MichaelWalker: ahh - the answer to me previous comment!
New question then - why would we ever want this to be null? Would we not want to fail fast if we couldn't create this?

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
adityapahuja: Policy, SAML SOAP Proxy, SAML Proxy configuration files have no EventEmitterConfiguration. The EventEmitterConfiguration will be null and Guice does not like having null value. That's why we use Optional.

hub/policy/src/test/java/uk/gov/ida/hub/policy/PolicyApplicationTest.java
adityapahuja: We do not want to fail fast. Instead, it will use stub queue by default when queue name is not provided i.e. event emitter will know that EventEmitterConfiguration is Optional.empty() and will use stub queue instead.

build.gradle
adityapahuja: rest-utils already has a dependency on event-emitter. When rest-utils no longer has dependency on event emitter, we can add event emitter to each component back (policy, saml soap proxy and saml proxy). It is not good idea to have verify-event-emitter in both verify-hub's gradle file and verify-utils-lib's gradle file as they can have different versions.

build.gradle
adityapahuja: Removed the hardcoded version.

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
MichaelWalker: How come they can need the EventEmitter without having a configuration - do they not need an 
 `EventEmitterConfiguration` in order to get an `EventEmitter`?

hub/policy/src/test/java/uk/gov/ida/hub/policy/PolicyApplicationTest.java
MichaelWalker: ah right. I see. I was kind of imagining still having a configuration, but the configuration explicitly saying it didn't have a queueName.

I generally dislike configuring things by the absence of stuff, but I think you are right that this probably works just fine by now.

build.gradle
MichaelWalker: That's a good and valid point, but I think I still slightly disagree.

In general, its bad practice to have hidden dependencies like this. I'm much happier when the gradle file is super explicit about the things that our application depends on.

This has a few advantages
- its clear and obvious what's going on
- we don't get any surprising failures if we were to remove `rest_utils`
- the version we use ought not be tightly coupled to the version used by `rest_utils` (if it is - that's a smell in itself) Ideally each library ought to be a nice, self-contained thing.

I take your point about not wanting the libs to get out of step, but I don't think its enough justification to not be open about our dependencies.
Worst case scenario if they do get out of step is a compile time error.

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
adityapahuja: If Event Emitter Configuration containing a source queue name is not provided, Event Emitter will use stub queue by default. Stub queue does not need source queue name and it writes the encrypted event message to a standard output.

build.gradle
adityapahuja: My previous developers I paired with in Verify team and I faced issues with having explicit version number of Verify library in the build.gradle where dependencies were using that library in the past. Having explicit version number of Verify library in the gradle file where its dependencies are using that library has the following issues.

1. More maintenance work: We need to specify a correct version number of a Verify library. This involves checking dependencies to ensure that they use the correct version number of the Verify library. If they do not, we have to update dependencies to use the correct version number of the Verify library.

2. Multiple versions of a library: When the version number of a Verify library is specified explicitly in the build.gradle file and somewhere in the dependencies, a dependency is using a different version number of the Verify library. We get multiple versions of the Verify library. Running compiling or executing integration test cases will return errors (compile-time error or runtime error).

3. Unseen impact after removing the library: If the explicit version number of library is removed from the build.gradle, we will not know which classes are affected by this change since dependencies are using the same library but with different version number.

MichaelWalker: Its true that dependency management in Java has its issues, but I don't think any of them are large enough to be worth hiding our dependencies. Hiding dependencies has all kinds of issues of its own (like making it very hard to figure out the impact of changing my lib (as I can't easily see where it is used) but more importantly it breaks much of the abstraction model of using libraries and as a result causes much pain. There might be individual cases where we could argue the trade off is worth it, but I don't agree that this is one of them.

My big issue is that it destroys the encapsulation of the `rest_utils` library.
Moving code to a library should come with 2 big advantages
1) reduces code duplication
2) abstracts away the implementation of a thing from the thing.

So if we have an example like the following
`A -> B -> C` (where by this I mean application `A` uses library `B` which in turn uses library `C`)

At the moment everything is happy. `A` uses `B` to do a thing, and doesn't care at all how `B` goes about doing it. This is really nice - we've successfully decoupled 'do the thing' from 'how to do the thing'

This means that `B` is totally free to change 'how to do the thing' as much as it likes, as long as it still retains the same interface for 'do the thing' `A` really doesn't care at all.

For example, if `B` were to decide that library `D` is actually better than `C` for some reason, we can just swap it out, to get the perfectly working `A -> B -> D`.

Now lets say that `A` actually wants to use `C` too.

Option 1) is to just use `C`. We 'get away' with this because `B` happens to use `C`, but its far from ideal, as we've broken all of that nice stuff that we discussed above. Now, `A` cares deeply about some of the precise implementation details of `B`. (ie we care about a specific version of a specific dependency)
Now `B` is no longer free to change its implementation. The change to use `D` rather than `C` now causes unexpected errors when we try to update `A`.

Option 2) is to directly import `C`. I think this is better for lots of reasons
- `B` retains its abstraction
- `A`'s dependencies are now more straight forward (they are nice and flat rather than nested, and everything is nice and explicit)
- `A` and `B` are now free to depend on different versions of `C` if they wish. - `C` is just an implementation detail of `B`, so as long as `B` is happy, `A` really shouldn't care what version of `C` `B` is using. (this is the part that Java handles quite poorly though - although I heard somewhere that it would be better in Java 9)
- As someone changing `C` I can now see that its being used by `A`

So..... I can imagine that there are situations where there might be a really compelling reason to not stick to the modular, encapsulated pattern, but I think we'd need a really powerful reason, as there are some really good reasons to stick with it.

In response to your points

1) I don't really agree. In either case in the examples above, updating `C` will require me to update `B` and update `A`. There might be one extra line of change in `A` with option 2, but I don't think its a big deal. We do have to pick a specific version, but that's a good thing. We should be caring about the version of the libraries we directly use.

2) I kind of agree. As in my explanation above, in principal I think its good for the versions to be able to change independently - they ought to be independent implementation details after all. (if that's not the case then that's a smell, and we should maybe refactor). I totally get that Java makes having multiple versions of a thing a pain though. We might be able to work around it with the `force` keyword, but you are totally right that Java doesn't play nicely with multiple versions.
I think the better fix would be to update `B` to have the latest version of `C` though.

3) I don't really understand this one - I'm arguing that you shouldn't ever have dependencies that you haven't explicitly declared in the gradle file, so you'd never remove them (unless you no longer needed the dependency - in which case everything is fine)



I hope that makes some kind of sense! Let me know if it doesn't. (and sorry for such a long comment)

Basically I understand your point, but I feel that we lose far more by making dependencies transitive than we gain.

adityapahuja: Regarding my third point, there are three libraries (`A`, `B` and `C`) and there are two versions of library `C(v1)` and `C(v2)`.

`A -> B -> C (v1)`
`|`
`---> C (v2)`

Suppose that 
* both libraries `C(v1)` and `C(v2)` have a same class `K` and the class `K` have a same method `M`
* `K(v1).M` method from the class `K` in the library `C(v1)` and `K(v2).M` method from the class `K` in library `C(v2)` do different things and produce different outcomes.
* classes in the library `A` are using `K(v2).M`.

If we remove the explicit library `C(v2)` from build.gradle in the library `A`, all classes in library `A` affected by this change will start using `K(v1).M` instead. They will compile successfully. They will fail in integration tests only if these tests exist to test classes in the library `A` using `K(v2).M`.

`A -> B -> C (v1)`

Another example is that if we have

`A -> B -> C (v2)`
`|`
`---> C(v2)`

If we remove the explicit library `C(v2)` from build.gradle in the library `A`, all classes in the library `A` affected by this change will still use `K(v2).M` indirectly via the library `B`.  They will compile and pass integration tests only if these tests exist to test classes in the library `A` using `K(v2).M`.


Suppose that both libraries A and B are using different versions of SAML extension libraries. Library A uses a latest version of SAML extension library which supports a new attribute in SAML message whereas B is using an old version of SAML extension library which does not support the new attribute in the SAML message.

`A -> B -> old version of SAML extension library`
`|`
`---> new version of SAML extension library`

If we run an application with these dependencies defined above, it would probably break down due to some code in library B not recognising the new attribute in the SAML message. This issue would have been avoided if both A and B are using the same version of SAML extension library.

MichaelWalker: Yep - this is indeed a pain point with our setup.
I think there are 2 answers, a theoretical one for a generic language, and a Java one.

Generic World
This is why using libraries to share domain objects often does more harm than good. (its why we want to kill `saml-domain-objects` for example) If the library 'does a thing' rather than 'provides' an object, then its just an implementation detail, is weakly coupled, and its not really an issue to have different versions (each thing defines the what it needs and uses that)

As you rightly point out, things become harder if you are using a lib to provide a shared domain object. We've tried to mitigate this issue by using an interface rather than a class, which makes this issue smaller - but in theory there could indeed be an issue here. (most often one caught by the compiler)
The only real solution is to make sure we update all all users of our lib when we change the model - which is made simpler if all users declare that they are users.

Java World
So the big difference here is that the JVM doesn't really allow for multiple versions of the same library to exist. Which version gets picked depends on the dependency management tool you pick, but in gradle the most recent version of the lib wins (unless you `force` the version). https://gist.github.com/cr7pt0gr4ph7/6cd6339eabc2056bc3e7

In some ways this solves most of your concerns, but on the flip side it means that just because you picked a version in your gradle.build file, you can't be sure that's the one you actually use, which is confusing. The encapsulation of libraries is also diminished. You win some and you lose some I guess.


So again I think we get back to the same position, which I think is 'Yep - dependency management in Java is a pain' but I still don't really think we gain anything by deliberately not declaring our dependencies (and we might even have added a warning to our build).
I do think we damage the maintainability of our codebase by relying on transitive dependencies though, for all of the reasons in the previous comment.

adityapahuja: As requested, I made the changes.

MichaelWalker: As a side point, `Dependency Hell` is just a generally super interesting problem that is 'solved' in different ways by different languages and package managers (eg Java + Maven/Gradle take quite a different approach to Node + npm + JS). Java often gets criticised for its solution, and npm is often praised, but as with everything, there are ups and downs to each approach.

If its not a thing you've had a chance to read up on before it might be worth it. Happy to point at some articles if you'd like!

PR #34, opened 2018-02-28
bau: Nicer way of guarding against null

richardTowers:


hub/saml-soap-proxy/src/main/java/uk/gov/ida/hub/samlsoapproxy/runnabletasks/ExecuteAttributeQueryRequest.java
```diff
@@ -96,16 +98,15 @@ private void validateRequestSignature(Element matchingServiceRequest, URI matchi
     private void validateResponseSignature(Element responseFromMatchingService) {
         Response response = elementToSamlResponseTransformer.apply(responseFromMatchingService);
         SamlValidationResponse signatureValidationResponse = matchingResponseSignatureValidator.validate(response, AttributeAuthorityDescriptor.DEFAULT_ELEMENT_NAME);
-        String message = hasStatusMessage(response.getStatus()) ? response.getStatus().getStatusMessage().getMessage() : "";
+        String message = Optional.ofNullable(response.getStatus())
```
MichaelWalker: is `response.getStatus()` actually a nullable thing here? Previously it looks like this would have thrown an NPE if the status was null.

If that's the case, then can we simplify it further to just 
```
String message = Optional.ofNullable(response.getStatus().getStatusMessage())
    .map(StatusMessage::getMessage)
    .orElse("");
```

richardTowers: Depends what you mean by "nullable" - it would be possible to construct a SAML message where that was not present, and therefore null. It's possible we're null checking it somewhere else, but I'm not sure.

In this case we're only looking at it to construct a logging message, so an NPE would be breaking the user journey for a log message, so I think the `.orElse("")` is the right behaviour in this case.

richardTowers: But... I've made a discovery about this code, so I'm doing a bigger refactor. Second PR incoming.

PR #35, opened 2018-03-02
refactor: use a `NullResponseAssertionSigner`

chrisholmes:
Saml Engine should not be able to sign assertions in Responses so it shouldn't be
using an Assertion signer. This should be updated in `verify-hub-saml`
but for now I'm quick fixing it here by using a NullObject version.

### DIFF
--- a/hub/saml-engine/src/main/java/uk/gov/ida/saml/core/transformers/outbound/decorators/NullResponseAssertionSigner.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/saml/core/transformers/outbound/decorators/NullResponseAssertionSigner.java
@@ -1,0 +1,14 @@
+package uk.gov.ida.saml.core.transformers.outbound.decorators;
+
+import org.opensaml.saml.saml2.core.Response;
+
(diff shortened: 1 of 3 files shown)


hub/saml-engine/src/main/java/uk/gov/ida/saml/core/transformers/outbound/decorators/NullResponseAssertionSigner.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.ida.saml.core.transformers.outbound.decorators;
+
+import org.opensaml.saml.saml2.core.Response;
+
+public class NullResponseAssertionSigner extends ResponseAssertionSigner {
```
richardTowers: Can we make this a `ThrowingResponseAssertionSigner` to make sure we don't end up using it accidentally?

kornysietsma:
bump - @richardTowers or @chrisholmes - is this something we need, or can it be deleted?

vixus0:
Fix this in hub-saml instead.

PR #38, opened 2018-03-15
bau: make sure metadata refresh tasks are available on admin port

willp-bl:
These seem to have been removed?

Pair: WillPa

### DIFF
--- a/hub/saml-proxy/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlproxy/MetadataRefreshTaskIntegrationTest.java
+++ b/hub/saml-proxy/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlproxy/MetadataRefreshTaskIntegrationTest.java
@@ -1,0 +1,60 @@
+package uk.gov.ida.integrationtest.hub.samlproxy;
+
+import helpers.JerseyClientConfigurationBuilder;
+import io.dropwizard.client.JerseyClientBuilder;
+import io.dropwizard.client.JerseyClientConfiguration;
+import io.dropwizard.util.Duration;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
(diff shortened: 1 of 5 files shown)


chrisholmes:
If this is a feature you feel is important it is probably worth introducing a test for this.

willp-bl:
saml-soap-proxy binds the task directly, but saml-engine and saml-proxy need to bind two tasks so I've done that manually.

Also added tests to make sure these tasks are there.

hub/saml-proxy/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlproxy/MetadataRefreshTaskIntegrationTest.java
```diff
@@ -0,0 +1,60 @@
...
+                .port(samlProxyAppRule.getAdminPort())
+                .build())
+                .request()
+                .post(Entity.text("refresh!"));
+        assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+    }
+
+    @Test
+    public void eidasConnectorMetadataRefreshTaskWorks() {
+        final Response response = client.target(UriBuilder.fromUri("http://localhost")
```
IreneLau-GDS: This is not the connector metadata. SAML engine doesn't read it's own metadata.

As of the time of this change, it is connecting to the stub country's metadata so it is refreshing the resolver that caches the stub country's metadata.

In the new world, which is about to happen within a day, it is using the signed trust anchor.

willp-bl: Ok, no point changing it as this can be removed at that time then.  

Will the trust anchor code have a refresh task?

willp-bl: 🎉

PR #44, opened 2018-04-05
TT-2117: Improve the backwards compatibility of State objects

richardTowers:
Policy has a data store where it stores the session state. This data is
stored in serialized form using Java's default Serializable behaviour
(see [The javadoc for Serializable](https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html)).

Occasionally we need to refactor these State objects, for example by:
- Renaming a class or field for clarity
- Changing the type of a field (e.g. a guava Optional<T> to a java Optional<T>)
- Removing fields which we don't need anymore
- Adding new fields to store new data

Because we're not specifying the serialVersionUID on our state classes
we're reliant on the JVM to calculate these numbers based on the
algorithm described [here](https://docs.oracle.com/javase/8/docs/platform/serialization/spec/class.html#a4100).
This means that even simple changes to objects cause the serialized
versions to break.

The [The javadoc for Serializable](https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html) say:

> ... it is strongly recommended that all serializable classes
> explicitly declare serialVersionUID values, since the default
> serialVersionUID computation is highly sensitive to class details that
> may vary depending on compiler implementations, and can thus result in
> unexpected InvalidClassExceptions during deserialization.

... so we should specify them explicitly. The serialVersionUIDs in this
commit were calculated on the un-modified classes, so they should be
consistent with the serialized objects from the last release.

To make future breaking changes to State classes more explicit we've
added a suite of DeserializationTests which attempt to deserialize a
serialized version of each of the States. These should fail if any
changes are made that prevent old versions from deserializing. There's a
comment on the test class describing what the developer should do.


rhowe-gds:
I wholeheartedly approve of this - if we're doing Java serialisation we should at least do it properly. +100 for the tests.

hub/policy/src/test/java/uk/gov/ida/hub/policy/domain/state/serialization/DeserializationTest.java
```diff
@@ -0,0 +1,296 @@
...
+ *
+ * If you can't make the changes in a non-breaking way you should:
+ * - Ensure that deploying the code will not cause an outage, or in extreme cases agree a planned outage with stakeholders
+ * - Increment the serialVersionUID on the class you're breaking
+ * - Update the serialized data in the test so the test covers the new format
+ *
+ * @see <a href="https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html">The javadoc for Serializable</a>
+ * @see <a href="https://stackoverflow.com/a/285809/1344760">Stackoverflow explanation of serialVersionUID</a>
+ */
+public class DeserializationTest {
```
chrisholmes: Is there possibly a test we can write that checks whether we're testing a State class against its deserialized form? Perhaps something narly with reflection?

richardTowers: I thought about doing that, yeah. It would be possible, but it gets a bit filthy-looking. You'd have to:

* Use reflection to get all of the non-abstract classes implementing `State`
* Use reflection to list all of the methods on DeserializationTest
* Check that there's a correctly named method for every state

As it is there's a risk someone will add a new state, and these tests will give us a false sense of security, so I think that does have some value. If I get a chance once I've finished my current task I'll have a look at that. I'd prefer it if it didn't block this PR from getting merged though.

willp-bl:
I've not had a full look through this PR but could you add a test that fails deserialization?

richardTowers:
@willp-bl - I could, but I'm not sure if it would have any value. Perhaps as documentation of what Java does when you try to deserialize something that doesn't match the class definition? But once written the test would just pass forever more, so nobody will ever look at it...

hub/policy/src/test/java/uk/gov/ida/hub/policy/domain/state/serialization/DeserializationTest.java
richardTowers: Turns out `Use reflection to get all of the non-abstract classes implementing State` is actually pretty hard, due to java. Not sure it's worth it.

richardTowers:
Thanks Chris - I'll start a hub deployment to make sure I'm on it in case this breaks something.

PR #46, opened 2018-04-09
firebreak: Use new signature validators

chrisholmes:
saml-metadata-bindings now provides more support for using Metadata to
validate signatures using features provided by OpenSAML.

We can use these features to validate signatures from the IDPs (companies + eidas schemes) as well as the Hub


chrisholmes:
I'm closing this in favour of #47 , #48 and #49

PR #49, opened 2018-04-10
firebreak: use new signature valdiator support in saml-engine

chrisholmes:
Uses the metadata backed Signature Trust Engine support that was added
to  saml-metadata-bindings. This should allow for a single way of
validating signatures from entities that are represented in metadata
across our codebase.


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
```diff
@@ -307,22 +302,24 @@ private MetadataHealthCheck getVerifyMetadataHealthCheck(
...
     @Provides
+    @Named("VERIFY_METADATA_REFRESH_TASK")
     @Singleton
-    private IdpSingleSignOnServiceHelper getIdpSingleSignOnServiceHelper(@Named("VerifyMetadataResolver") MetadataResolver metadataResolver) {
-        return new IdpSingleSignOnServiceHelper(metadataResolver);
+    private Task registerMetadataRefreshTask(Environment environment, @Named(VERIFY_METADATA_RESOLVER) MetadataResolver metadataResolver) {
+        Task task = new Task("metadata-refresh") {
+            @Override
+            public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
+                ((AbstractReloadingMetadataResolver) metadataResolver).refresh();
```
richardTowers: Is it possible to avoid this cast by changing the parameter's type? Feels a bit weird to have a method that takes a `MetadataResolver` and only works properly if you pass the right subtype in.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineApplication.java
```diff
@@ -41,8 +45,13 @@
 
 public class SamlEngineApplication extends Application<SamlEngineConfiguration> {
 
+    private final MetadataResolverBundle<SamlEngineConfiguration> verifyMetadataBundle;
     private GuiceBundle<SamlEngineConfiguration> guiceBundle;
 
+    public SamlEngineApplication() {
+        verifyMetadataBundle = new MetadataResolverBundle<>((SamlEngineConfiguration::getMetadataConfiguration));
```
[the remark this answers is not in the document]
chrisholmes: I think it's just an IDE refactoring artifacts

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
chrisholmes: Part of the problem is that our factories for `MetadataResolver` only return a `MetadataResolver` rather than the type of what actually gets build `JerseyClientMetadataResolver` so we have to do this casting here. We'd have to change saml-lib to fix it.

At some point soon I'd like to move this binding behaviour into the `MetadataResolverBundle` where the refresh task will be added automatically without guice.

richardTowers: Ok, that's fair enough then.

PR #50, opened 2018-04-13
BAU: Adding 'null' as a supported version

smford:
HMRC customises their MSA which means that the version it returns to Verify is an unexpected 'null'.

This change adds 'null' to be an acceptable version and stops sensu errors occuring.

Pair: @smford & @willpa


richardTowers:
This problem is fixed in the latest MSA release (not saying we shouldn't work around it in the meantime, but just as an FYI).

PR #68, opened 2018-05-18
Log useful request / response attrs as JSON

davbo:
Currently the ProtectiveMonitoringLogFormatter logs something which at
first glance may appear to be JSON. It has {} and keys/values separated
by : however don't be fooled. This is not JSON, there are no quotes "!.

This change places attributes which were part of the LogFormatter into
the MDC which is included in the Logstash output as JSON. Having
structured logs makes it easier to query them through ELK.

As this change doesn't alter the message logged it shouldn't impact any
current users of the logs (e.g. SecOps).

### DIFF
--- a/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/logging/ProtectiveMonitoringLogger.java
+++ b/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/logging/ProtectiveMonitoringLogger.java
@@ -1,12 +1,18 @@
 package uk.gov.ida.hub.samlproxy.logging;
 
+import com.google.common.collect.ImmutableMap;
 import org.opensaml.saml.saml2.core.AuthnRequest;
+import org.opensaml.saml.saml2.core.Issuer;
 import org.opensaml.saml.saml2.core.Response;
+import org.opensaml.saml.saml2.core.StatusCode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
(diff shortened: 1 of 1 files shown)


davbo:
Struggled getting some of the tests passing (regardless of my changes) 
```
    CountryMetadataConsumerTest. shouldServeCountryMetadata
    MetadataConsumerTests. shouldAllowRequestsWhenMetadataIsAvailableAndValid
    SamlMessageReceiverApiResourceEidasEnabledTest. eidasResponsePost_shouldRespondWithSuccessWhenPolicyRespondsWithSuccess
    SamlMessageReceiverApiResourceTest. shouldCreateSessionForAuthnRequest
    SamlMessageReceiverApiResourceTest. shouldErrorWhenAProblemOccursWithinSessionProxy
    SamlMessageReceiverApiResourceTest. shouldErrorWhenAuthnRequestIsInvalid
```
Would appreciate some help getting these to work - or at least to validate that my change doesn't make them fail :-)

davbo:
This change places useful data (e.g. entity ID of IdP's) outside of the "message" field, making it possible for different types of queries upon our data in ElasticSearch. Data held in the "message" field doesn't have so-called "[fielddata](https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html)" associated. This is because "message" has such high cardinality.

This makes it impossible to perform aggregations within Elastic such as "count AuthN request by IdP". 

Tools such as Grafana use aggregations when [templating dashboards](http://docs.grafana.org/features/datasources/elasticsearch/#templating). This is helpful as the dashboard doesn't need to be edited as EntityIds change or new ones are added.

PR #69, opened 2018-05-18
VCU-43: Configurable SHA Signing Algorithm

dbes-gds:
Update the Hub (saml-engine) to sign with SHA256 by default.
Configuration option provided to revert to SHA1 if required.

Co-authored-by Alan Carter <alan.carter@digital.cabinet-office.gov.uk>


dbes-gds:
Closing while required saml-engine config is added.

PR #70, opened 2018-05-29
Handle cases where validSignature is null

davbo:
In some cases the Hub will not sign Responses (relying on signed
assertions) and this logging would error with a NullPointerException.

This change separates the states of not having a signature and that
signature not being valid to better represent what is happening.

### DIFF
--- a/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/controllogic/SamlMessageSenderHandler.java
+++ b/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/controllogic/SamlMessageSenderHandler.java
@@ -12,6 +12,7 @@ import uk.gov.ida.hub.samlproxy.logging.ExternalCommunicationEventLogger;
 import uk.gov.ida.hub.samlproxy.logging.ProtectiveMonitoringLogger;
 import uk.gov.ida.hub.samlproxy.proxy.SessionProxy;
 import uk.gov.ida.hub.samlproxy.repositories.Direction;
+import uk.gov.ida.hub.samlproxy.repositories.SignatureStatus;
 import uk.gov.ida.saml.core.validation.SamlTransformationErrorException;
 import uk.gov.ida.saml.core.validation.SamlValidationResponse;
 import uk.gov.ida.saml.core.validation.SamlValidationSpecificationFailure;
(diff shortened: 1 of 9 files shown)


IreneLau-GDS:
Actually, if you look at the usages of that method, you'll see that when it's called by ```uk.gov.ida.hub.samlproxy.controllogic. SamlMessageSenderHandler :: validateAndLogSamlResponseSignature ```(https://github.com/alphagov/verify-hub/blob/master/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/controllogic/SamlMessageSenderHandler.java#L126)

You'll see that the 3rd arg means if there is an issuer or not.

I think, in the case when the RP is of "simple profile", this mean issuer will be null.

willp-bl:
Might be better to change logAuthnRequest to take an optional as the third argument - would make the logic read a bit better in the caller and the handling of the param a bit nicer - https://github.com/alphagov/verify-hub/blob/91561de04cc2e74ce42641de7bb046d06c9820d4/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/controllogic/SamlMessageSenderHandler.java#L137 would be `empty()` instead of `null`

davbo:
@willp-bl I think Optional<T> of Java's boxed primitives creates issues?

willp-bl:
Boolean foo = null;
Optional bar = Optional.ofNullable(foo);
bar.isPresent() == false

I feel like I am missing something by your response though :)

davbo:
Yeah - I thought `Optional<T>` types cause issues with primitive boxed types. That's why they have a separate `OptionalInt` that said I can't see an `OptionalBoolean` so I guess it's fine.

Just looked it up in Effective Java and it seems to be performance concerns about having 2 levels of boxing. Doesn't sound like something we need to be concerned about? If you agree I'll refactor to use Optional.

willp-bl:
I don't think this will be a performance bottleneck :) Plus the compiler/JVM should work out the most efficient code path.

davbo:
@willp-bl rather than introduce `Optional<T>` I'd sooner remove the boxed type all-together. What do you think to https://github.com/alphagov/verify-hub/pull/71 ?

richardTowers:
💯 from me for #71 - using `Boolean` to represent three possible states (true, false, null) feels super hacky. Either two booleans or an enum would be clearer.

Re: using Optional<> for the argument - two different objections to David's thing about boxing:
- Same problem as `Boolean` with three states, only now you've got even more (`of(true), of(false), empty(), null`)
- Optional is only intended for return types, not method args, as discussed here - https://stackoverflow.com/questions/23454952/uses-for-optional/23464794#23464794

IreneLau-GDS:
I have looked at the latest (after #71 is merged into this one). The changes makes a lot of sense now!

Unfortunately I have meetings lined up for the next two hours so I don't want to merge this now and then run away with no way to monitor the hub and joint acceptance Jenkins build.

Can someone kindly merge it?

IreneLau-GDS:
I'm out of the meeting now. Will merge this now!

PR #71, opened 2018-05-29
Remove boxed Boolean type from ProtMonLogging

davbo:
Previously we were using a Boolean type to represent the following
 - null: no signature on the request/response
 - false: signature present but not valid
 - true: signature present and valid

This can be problematic as null checks need to be carefully performed.
These checks were missing in a035bb51a2f469022dbb25bd163e4d9a49d3ce8b

This commit replaces the usage of `null` with a separate isSigned
boolean. The change is only necessary for when processing Response
objects as they have this option of being unsigned (simple profile).


richardTowers:
Not a major concern, but using two booleans here means we've got four possible states, one of which doesn't make sense:

* `isSigned && signatureIsValid`
* `isSigned && !signatureIsValid`
* `!isSigned && !signatureIsValid`
* `!isSigned && signatureIsValid // Doesn't make sense`

You could avoid this (at the cost of adding another type) by adding an enum like `enum { VALID_SIGNATURE, INVALID_SIGNATURE, NO_SIGNATURE }`, which I think might be clearer and avoids the nonsense scenario.

IreneLau-GDS:
I like the new enum way a lot. 

Any further comments anyone? @willp-bl ? 

eIDAS has some urgent fixes that need to go but we need this set of changes to go first or hub will be unhappy.

willp-bl:
LGTM

willp-bl:
(we'll need to look at the eidas control flow after)

davbo:
This seems to be the preferred to having the boxed Boolean's so I'll merge it into #70

PR #74, opened 2018-05-31
BAU: Updated stub-idp-saml library to the latest version

CharlesIC:


andy-paine:
The fact you have to do this in hub upsets me surprisingly deeply.

PR #76, opened 2018-06-05
VCU-43: Configurable Signing Algo Per Service: Config Changes

dbes-gds:
Change the Config microservice to return the signing algo flag value from the transaction config to the saml-engine service.
Plug-in to saml-engine.


dbes-gds:
Yes, as advised by Codacy.  It rejected them in part 1 so in part 2 I
updated them beforehand.

alan-gds:
The dev team has agreed that we can ignore this Codacy check, and Dan has reinstated the underscores; I'll merge the PR.

PR #83, opened 2018-06-19
bau: indirection minus minus

willp-bl:
Pair: WillPa


richardTowers:
But what about compat with Java 4 :trollface: ?

PR #89, opened 2018-07-12
BAU: Import configurations for policy, config and saml-soap-proxy

rhowe-gds:
These services have generic configurations which do not differ between
environments, save for values sourced from environment variables. Move them
into this repository and the service packages.


debian/saml-soap-proxy/saml-soap-proxy.yml
```diff
@@ -0,0 +1,162 @@
...
+      - type: logstash-file
+        currentLogFilename: /var/log/ida/saml-soap-proxy_access.log
+        archivedLogFilenamePattern: /var/log/ida/saml-soap-proxy_access.log.%d.gz
+        archivedFileCount: 7
+      - type: logstash-syslog
+        facility: local2
+        tag: saml-soap-proxy
+
+saml:
+  entityId: https://signin.service.gov.uk
```
JHjava: Should the entityId and the eidasEntityId be environment variables?

willp-bl: Yes

debian/saml-soap-proxy/saml-soap-proxy.yml
```diff
@@ -0,0 +1,162 @@
...
+      archivedFileCount: 7
+    - type: logstash-syslog
+      facility: local2
+      tag: saml-soap-proxy
+    - type: raven
+      dsn: ${SENTRY_DSN:-naive+https://sentry/1}
+      threshold: ERROR
+      tags: service-name:saml-soap-proxy
+
+metadata:
```
JHjava: The signin.service of the uri and the expectedEntityId should also be environment variables.

debian/saml-soap-proxy/saml-soap-proxy.yml
rhowe-gds: Yes, I was clearly not paying enough attention when putting this PR together. This is an old version of the configuration. I've redone the PR now with the correct config file.

Good spot @JHjava

debian/saml-soap-proxy/saml-soap-proxy.yml
rhowe-gds: Fixed. I'd brought in a stale version of the config - whoops! Top spot!

willp-bl:
Can it be enabled for all environments now? Stub-country already exists everywhere

rhowe-gds:
Yes, something like that. Currently the only differences in those apps' config are the eIDAS configuration section. If that difference can be removed, we can move those configurations into this repo as well.

rhowe-gds:
I know that difference can't be removed currently, but if we can find a way to make it not be different (be it an enable flag or shipping eidas config everywhere or something else) then we can move those configurations into this repository. I wasn't saying to just delete the differences - obviously that doesn't fly

PR #91, opened 2018-07-18
BAU: Upgrade DW (and other libs) + remove stub-idp-saml

andy-paine:
- Upgrade dropwizard to 1.3.5
- Upgrade other libs to latest
- Move DW version into ext so that dropwizard and
dropwizard-infinispan are pegged to same base version
- Remove stub-IDP-saml by importing some methods and using new saml-test-utils


build.gradle
```diff
@@ -24,19 +24,20 @@ apply plugin: 'com.github.ben-manes.versions'
...
 
 def dependencyVersions = [
-            ida_utils:'336',
-            dropwizard:'1.1.4',
-            dropwizard_infinispan:'1.1.4-45',
+            ida_utils:'337',
+            dropwizard:"$dropwizard_version",
+            dropwizard_infinispan:"$dropwizard_version-47",
             pact:'3.5.6',
             stub_idp_saml:"$opensaml_version-72",
```
adityapahuja: Please update stub idp saml to use latest version of saml libs and ida utils.

build.gradle
```diff
@@ -102,7 +103,7 @@ subprojects {
                 'io.dropwizard:dropwizard-auth:' + dependencyVersions.dropwizard,
                 'io.dropwizard:dropwizard-configuration:' + dependencyVersions.dropwizard,
                 'io.dropwizard:dropwizard-metrics-graphite:' + dependencyVersions.dropwizard,
-                'com.hubspot.dropwizard:dropwizard-guicier:0.9.1.0',
+                'com.hubspot.dropwizard:dropwizard-guicier:1.0.0.6',
```
adityapahuja: Please update dropwizard guicer to the latest version '1.0.9.1'.

andy-paine: Silly me for assuming their `releases` page contained their releases https://github.com/HubSpot/dropwizard-guicier/releases !

andy-paine: 1.0.9.1 has build errors and it is outside the scope of this work to do the full upgrade. I will raise it as a tech-debt card for the relevant team

adityapahuja: There is 'com.hubspot.dropwizard:dropwizard-guicier:0.9.1.0' in the force section. Removing it breaks hub.

adityapahuja:
I ran pre-commit.sh and it failed due to config failure to start.

adityapahuja:
Second, third, fourth and fifth commit fail pre-commit.sh. We should not have commits which fail pre-commit.sh in master.

andy-paine:
@adityapahuja The config failure as discussed is not within the scope of this work. The interim commits failing is not an issue - if we require every commit to pass then we lose the benefit of making small commits. Given this is being lumped into a single PR, it is sufficient that it passes at the beginning and end of the PR not necessarily in the middle.

PR #95, opened 2018-07-30
BAU: Add use of enabled flag

andy-paine:
- Joint & Integration now have an "enabled" flag in their country
configuration section
- This will now read the value of this
- After this is released we can add the "country" section to all
env configs and just have "enabled" set to false
- Eventually we can make the "country" section mandatory and only
use the enabled flag for deciding whether to activate eIDAS


JHjava:
Do we need to add the enabled flag to the country in the local config?

andy-paine:
Right you are. Incoming.

hub/saml-proxy/src/test/resources/saml-proxy.yml
```diff
@@ -95,25 +95,27 @@ metadata:
       verifyHostname: false
       trustSelfSignedCertificates: true
 
-eidas: true
-countryMetadata:
```
andy-paine: Good thing `ignoreUnknownProperties` was on and so this was completely ignored 😱

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineConfiguration.java
```diff
@@ -111,11 +111,11 @@ public SamlConfiguration getSamlConfiguration() {
     }
 
     public Optional<CountryConfiguration> getCountryConfiguration() {
-        return country != null ? Optional.of(country) : Optional.empty();
+        return Optional.ofNullable(country);
     }
 
     public boolean isEidasEnabled() {
-        return getCountryConfiguration().isPresent();
+        return getCountryConfiguration().isPresent() && getCountryConfiguration().get().isEnabled();
```
FelixMarcusMillne: _small innocuous point_ Optional best practice would be something like:

`return getCountryConfiguration().map(CountryConfiguration::isEnabled).orElse(false)`

andy-paine: Fair point - however I'm planning on removing the Optional and just doing `getCountryConfiguration.isEnabled` next release :D I figured this was easier to review the new behaviour

PR #99, opened 2018-08-03
EID-817 - Give eidas support for required algorithms

JHjava:
- Give eidas flow support to the encryption algorithms for content
encryption and key transport as in the eidas crypto spec. 
- Renamed method Named attribute in SamlEngineModule as it is only used for Eidas and doesn't just use AES256 anymore.
- The best place to test this would be from StubCountry, encrypting the assertion using the algorithms in the crypto spec. This would give full end to end coverage. 
- This change will cover EID-817 and EID-820

### DIFF
--- a/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
@@ -248,7 +248,7 @@ public class SamlEngineModule extends AbstractModule {
                                                                                            IdpIdaStatusUnmarshaller idpIdaStatusUnmarshaller,
                                                                                            @Named("ResponseAssertionsFromCountryValidator") Optional<ResponseAssertionsFromCountryValidator> responseAssertionFromCountryValidator,
                                                                                            Optional<DestinationValidator> validateSamlResponseIssuedByIdpDestination,
-                                                                                           @Named("AES256DecrypterWithGCM") AssertionDecrypter assertionDecrypter,
+                                                                                           @Named("EidasAssertionDecrypter") AssertionDecrypter assertionDecrypter,
                                                                                            AssertionBlobEncrypter assertionBlobEncrypter,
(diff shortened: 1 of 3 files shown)


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
```diff
@@ -643,12 +643,18 @@ private DecoratedSamlResponseToIdaResponseIssuedByIdpTransformer getResponseToIn
     }
 
     @Provides
-    @Named("AES256DecrypterWithGCM")
+    @Named("EidasAssertionDecrypter")
     private AssertionDecrypter getAES256WithGCMAssertionDecrypter(IdaKeyStore keyStore) {
```
vixus0: Might want to rename this method.

JHjava:
No, because it is only used for the CountryAuthnResponseTranslatorService. The normal flow AssertionDecrypter is created in the HubTransformerFactory under the method getSamlResponseAssertionDecrypter and uses the EncryptionAlgorithmValidator no argument constructor.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
JHjava: Good spot. Will make this change to bring in inline with the @Named name.

IreneLau-GDS:
Where are the tests? 🤔

JHjava:
We're just changing the values which we are passing to the constructor in saml-libs so not sure what tests would have been appropriate or if they would have added value? Open to suggestions. If we encrypt the algorithms in Stub Country as required, then any problems should be picked up in the end to end tests.

vixus0:
@IreneLau-GDS the `EncryptionAlgorithmValidator`'s whitelisting powers are unit tested in saml-security, so I don't think that needs to happen here.

IreneLau-GDS:
Man... It always makes me cringe when we have to main the whitelist in the test descrypter too. But then again it might not be such a bad idea.

PR #103, opened 2018-08-07
EID-820 - Use correct algorithm when decrypting eidas assertions

JHjava:
- Was previously AES256-CBC, should have been AES128-GCM as in the spec.
- Create tests to generate Saml Response using different encryption algorithms,
so we can ensure that Hub can handle them.

### DIFF
--- a/hub/saml-engine/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlengine/apprule/CountryAuthnResponseTranslatorResourceTest.java
+++ b/hub/saml-engine/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlengine/apprule/CountryAuthnResponseTranslatorResourceTest.java
@@ -117,6 +117,78 @@ public class CountryAuthnResponseTranslatorResourceTest {
         assertThat(response.getStatus()).isEqualTo(Response.Status.BAD_REQUEST.getStatusCode());
     }
 
+    @Test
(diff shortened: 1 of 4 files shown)


hub/saml-engine/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlengine/apprule/CountryAuthnResponseTranslatorResourceTest.java
```diff
@@ -135,22 +207,28 @@ private void assertEquals(Assertion first, Assertion second) {
     }
 
     private SamlAuthnResponseTranslatorDto createAuthnResponseSignedByKeyPair(String publicKey, String privateKey) throws Exception {
+        return createAuthnResponseSignedByKeyPair(publicKey, privateKey, EncryptionConstants.ALGO_ID_BLOCKCIPHER_AES256_GCM, EncryptionConstants.ALGO_ID_KEYTRANSPORT_RSAOAEP);
```
IreneLau-GDS: I'm going to be pedantic... 

This means we are only testing responses of AES256_GCM but we are accepting AES128_GCM as well (I can see it in the module class.)

Should we try to cover both cases?

IreneLau-GDS: Okay I was being daft. Thx for pointing it out in person.

PR #104, opened 2018-08-07
EID-842: Accept transient NameIDs

vixus0:
The German eIDAS Middleware returns NameIDs with the format set to
*transient*. Normally, Verify IDPs return *persistent*. We should allow
the validator to accept both types of NameID since it doesn't really
affect the journey.


hub-saml/src/main/java/uk/gov/ida/saml/core/validators/subject/AssertionSubjectValidator.java
```diff
@@ -26,7 +29,11 @@ public void validate(
             throw new SamlTransformationErrorException(failure.getErrorMessage(), failure.getLogLevel());
         }
 
-        if (!NameIDType.PERSISTENT.equals(subject.getNameID().getFormat())) {
+        boolean correctNameIdType = Stream
+                .of(NameIDType.PERSISTENT, NameIDType.TRANSIENT)
+                .anyMatch(type -> type.equalsIgnoreCase(subject.getNameID().getFormat()));
+
```
kerrr: Should we ignore case here? URIs are case sensitive and we've pushed back on this before. Do we need this for what DE are sending?

PR #109, opened 2018-08-13
BAU: Make session store usage more general

andy-paine:
- I would recommend reviewing this on a per-commit basis
- Interesting things to note:
 * The tests in TransitionCacheTests (you should convince yourself that this is the correct behaviour for a read/write + write only cache)
 * This is essentially a no-op for the existing system (tests have only changed to handle new config format)
- The context here is that we would like to be able to use a second cache, either to support the transition to another serialization format or to another cache provider

Test diff: `2 files changed, 520 insertions(+), 8 deletions(-)` 🎉 
Non-test diff: ` 16 files changed, 394 insertions(+), 75 deletions(-)`

### DIFF
--- a/configuration/local/policy.yml
+++ b/configuration/local/policy.yml
@@ -14,14 +14,16 @@ logging:
   appenders:
     - type: console
 
-
-infinispan:
-  bindAddress: 
-  initialHosts: 
-  clusterName: 
-  type: standalone
-  expiration: 8h
-  persistenceToFileEnabled: false
+sessionStoreConfiguration:
+  infinispanConfiguration:
+    enabled: true
+    configuration:
+      bindAddress:
+      initialHosts:
+      clusterName:
+      type: standalone
+      expiration: 8h
+      persistenceToFileEnabled: false
 
 eventSinkUri: ${EVENT_SINK_URL}
 
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyApplication.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyApplication.java
@@ -1,8 +1,6 @@
 package uk.gov.ida.hub.policy;
 
 import com.fasterxml.jackson.databind.util.ISO8601DateFormat;
-import com.google.inject.AbstractModule;
-import com.google.inject.Module;
 import com.hubspot.dropwizard.guicier.GuiceBundle;
 import engineering.reliability.gds.metrics.bundle.PrometheusBundle;
 import io.dropwizard.Application;
@@ -14,6 +12,8 @@ import uk.gov.ida.bundles.LoggingBundle;
 import uk.gov.ida.bundles.MonitoringBundle;
 import uk.gov.ida.bundles.ServiceStatusBundle;
 import uk.gov.ida.eventemitter.EventEmitterModule;
+import uk.gov.ida.hub.policy.domain.SessionId;
+import uk.gov.ida.hub.policy.domain.State;
 import uk.gov.ida.hub.policy.domain.exception.SessionAlreadyExistingExceptionMapper;
 import uk.gov.ida.hub.policy.domain.exception.SessionCreationFailureExceptionMapper;
 import uk.gov.ida.hub.policy.domain.exception.SessionNotFoundExceptionMapper;
@@ -38,6 +38,7 @@ import uk.gov.ida.shared.dropwizard.infinispan.util.InfinispanBundle;
 import uk.gov.ida.shared.dropwizard.infinispan.util.InfinispanCacheManager;
 
 import javax.inject.Provider;
+import java.util.concurrent.ConcurrentMap;
 
 public class PolicyApplication extends Application<PolicyConfiguration> {
 
@@ -66,17 +67,30 @@ public class PolicyApplication extends Application<PolicyConfiguration> {
         bootstrap.addBundle(new LoggingBundle());
         bootstrap.addBundle(new PrometheusBundle());
         bootstrap.addBundle(new IdaJsonProcessingExceptionMapperBundle());
+
         final InfinispanBundle infinispanBundle = new InfinispanBundle();
-        // the infinispan cache manager needs to be lazy loaded because it is not initialized at this point.
         bootstrap.addBundle(infinispanBundle);
+
+        final SessionBundle sessionBundle = new SessionBundle(infinispanBundle.getInfinispanCacheManagerProvider());
+
+        // the infinispan cache manager needs to be lazy loaded because it is not initialized at this point.
+        bootstrap.addBundle(sessionBundle);
         guiceBundle = GuiceBundle.defaultBuilder(PolicyConfiguration.class)
-                .modules(getPolicyModule(), new EventEmitterModule(),  bindInfinispan(infinispanBundle.getInfinispanCacheManagerProvider()))
+                .modules(getPolicyModule(),
+                        new EventEmitterModule(),
+                        getSessionModule(infinispanBundle.getInfinispanCacheManagerProvider(),
+                                sessionBundle.getSessionStateStoreProvider()))
                 .build();
         bootstrap.addBundle(guiceBundle);
     }
 
     protected PolicyModule getPolicyModule() {
         return new PolicyModule();
+    }
+
+    protected SessionModule getSessionModule(Provider<InfinispanCacheManager> infinispanCacheManagerProvider,
+                                             Provider<ConcurrentMap<SessionId, State>> sessionStateStoreProvider) {
+        return new SessionModule(infinispanCacheManagerProvider, sessionStateStoreProvider);
     }
 
     @Override
@@ -111,14 +125,5 @@ public class PolicyApplication extends Application<PolicyConfiguration> {
             environment.jersey().register(CountriesResource.class);
             environment.jersey().register(EidasSessionResource.class);
         }
-    }
-
-    protected Module bindInfinispan(Provider<InfinispanCacheManager> cacheManagerProvider) {
-        return new AbstractModule() {
-            @Override
-            protected void configure() {
-                bind(InfinispanCacheManager.class).toProvider(cacheManagerProvider);
-            }
-        };
     }
 }
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyConfiguration.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyConfiguration.java
@@ -8,6 +8,7 @@ import io.dropwizard.client.JerseyClientConfiguration;
 import io.dropwizard.util.Duration;
 import uk.gov.ida.common.ServiceInfoConfiguration;
 import uk.gov.ida.configuration.ServiceNameConfiguration;
+import uk.gov.ida.hub.policy.configuration.SessionStoreConfiguration;
 import uk.gov.ida.restclient.RestfulClientConfiguration;
 import uk.gov.ida.shared.dropwizard.infinispan.config.InfinispanConfiguration;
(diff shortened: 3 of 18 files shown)


hub/policy/src/test/java/uk/gov/ida/hub/policy/controllogic/TransitionCacheTest.java
```diff
@@ -0,0 +1,407 @@
...
+
+public class TransitionCacheTest {
+
+    @Test
+    public void sizeShouldReturnSizeOfExisting() {
+        ConcurrentMap existing = new ConcurrentHashMap() {{
+            put(new SessionId("key1"), "value1");
+            put(new SessionId("key2"), "value2");
+        }};
+        ConcurrentMap transition = new ConcurrentHashMap() {{
```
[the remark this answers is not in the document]
andy-paine: It's a test

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyModule.java
```diff
@@ -124,14 +124,16 @@ public JsonClient forSamlSoapProxy(JsonResponseProcessor responseProcessor, Poli
...
+    public ConcurrentMap<SessionId, State> sessionCache(
+            InfinispanCacheManager infinispanCacheManager) {
+        return CacheFactory.getStateCache(infinispanCacheManager);
     }
 
     @Provides
     @Singleton
-    public ConcurrentMap<SessionId, DateTime> datetime_cache(InfinispanCacheManager infinispanCacheManager) {
-        return infinispanCacheManager.getCache("datetime_cache");
+    public ConcurrentMap<SessionId, DateTime> datetime_cache(
```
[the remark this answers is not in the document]
andy-paine: I merely edited this

configuration/local/policy.yml
```diff
@@ -14,14 +14,16 @@ logging:
...
 
-
-infinispan:
-  bindAddress: 
-  initialHosts: 
-  clusterName: 
-  type: standalone
-  expiration: 8h
-  persistenceToFileEnabled: false
+cache:
```
thomasalanlee: A general point but I'm not sure we should call infinispan a cache. We don't use it to increase access speed to a backing store so its more just a 'store' or a 'repository'

hub/policy/src/main/java/uk/gov/ida/hub/policy/factories/CacheFactory.java
```diff
@@ -0,0 +1,27 @@
...
+import uk.gov.ida.hub.policy.controllogic.TransitionCache;
+import uk.gov.ida.hub.policy.domain.SessionId;
+import uk.gov.ida.hub.policy.domain.State;
+import uk.gov.ida.shared.dropwizard.infinispan.util.InfinispanCacheManager;
+
+import java.util.Optional;
+import java.util.concurrent.ConcurrentMap;
+
+public class CacheFactory {
+
```
thomasalanlee: We could refactor this so the getStateCache() and getDateTimeCache becomes a single method get(String cachename). We could also replace the strings with an enum to prevent invalid cache names being passed.

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyConfiguration.java
```diff
@@ -157,4 +158,9 @@ public boolean isEidasEnabled() {
     public EventEmitterConfiguration getEventEmitterConfiguration() {
         return eventEmitterConfiguration;
     }
+
+    @Override
+    public InfinispanConfiguration getInfinispan() {
```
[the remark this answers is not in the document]
andy-paine: This is implementing `getInfinispan()` from the `InfinispanServiceConfiguration` interface (which is from one of our libs) so I don't really have any control over the naming 😞See [dropwizard-infinispan](https://github.com/alphagov/dropwizard-infinispan/blob/master/src/main/java/uk/gov/ida/shared/dropwizard/infinispan/config/InfinispanServiceConfiguration.java) for the poor naming

andy-paine:
@willp-bl @thomasalanlee I've updated based on suggestions.
Points of note:
* Changed (generally) from cache -> store however all the infinispan lib stuff still refers to it as cache
* Named the two concurrent maps to `SessionExpirationStore` and `SessionStateStore`
* Added some braces @willp-bl 😉- no more [goto fail](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/)
* Extracted the method in the `SessionStoreFactory` (previously `CacheFactory`)

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyConfiguration.java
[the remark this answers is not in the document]
willp-bl: fair enough - could be a good change to make in the lib if possible

andy-paine:
I've added a new commit related to metrics/healthchecking the 2 session stores. Noteworthy points:
- I only bothered to do the State store as the datetime store is hopefully going away
- This _does_ log session IDs in the healthcheck but I don't consider these sensitive as it is on the admin endpoint and these things are already logged
- I don't really understand Guice + DropWizard all that well so what I've done with the bundle stuff might be absolutely horrendous but it works on my machine
- Our integration tests for Policy are worringly brittle due to using stuff like `PolicyIntegrationApplication`

When we actually run this, we should be able to see a nice graph of the `transition` cache slowly approaching the `existing` cache count before exactly tracking each other.

willp-bl:
I don't think that the figures will ever exactly match due to the expiry times being handled separately by the respective stores/caches

hub/policy/src/test/java/uk/gov/ida/hub/policy/controllogic/TransitionStoreTest.java
```diff
@@ -0,0 +1,408 @@
...
+        SessionId key2 = new SessionId("key2");
+        Map putAll = new HashMap() {{
+            put(key2, "value2");
+            put(key3, "value3");
+        }};
+
+        TransitionStore transitionStore = new TransitionStore(existing, Optional.of(transition));
+        transitionStore.putAll(putAll);
+        assertThat(existing.containsKey(key2) && existing.containsKey(key3)).isTrue();
+        assertThat(existing.containsKey(key2) && existing.containsKey(key3)).isTrue();
```
adityapahuja: Can this line be `assertThat(transition.containsKey(key2) && existing.containsKey(key3)).isTrue();`?

hub/policy/src/test/java/uk/gov/ida/hub/policy/controllogic/TransitionStoreTest.java
```diff
@@ -0,0 +1,408 @@
...
+    }
```
adityapahuja: Can this line be assertThat(transition.containsKey(key2) && transition.containsKey(key3)).isTrue();?

hub/policy/src/test/java/uk/gov/ida/hub/policy/controllogic/TransitionStoreTest.java
```diff
@@ -0,0 +1,510 @@
...
+            put(new SessionId("key3"), "value3");
+        }};
+
+        TransitionStore transitionStore = new TransitionStore(existing, Optional.of(transition));
+        assertThat(transitionStore.transitionSize()).isEqualTo(transition.size());
+    }
+
+    @Test
+    public void getDifferingKeysShouldGiveSessionIdsOnlyInExistingMap() {
+        SessionId key1 = new SessionId("key1");
```
adityapahuja: getDifferingSessionIds() will return empty if both existing and transition have same session ids but these session ids have different values. Can we add a test case to test this scenario?

configuration/local/policy.yml
adityapahuja: I see `cache`  has been replaced with `sessionStoreConfiguration`.  👍

hub/policy/src/main/java/uk/gov/ida/hub/policy/factories/SessionStoreFactory.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public static TransitionStore<State> getSessionStateStore(
+            InfinispanCacheManager infinispanCacheManager
+    ) {
+        return getSessionStore(infinispanCacheManager, "state_cache");
+    }
+
+    public static ConcurrentMap<SessionId, DateTime> getSessionExpirationStore(
+            InfinispanCacheManager infinispanCacheManager
+    ) {
```
adityapahuja: Can we have a more general method `getSessionStore(InfinispanCacheManager, CacheName)` instead of having two specialised methods `getSessionStateStore(InfinispanCacheManager)` and `getSessionExpirationStore(InfinispanCacheManager)` and replace hardcoded values `state_cache` and `datetime_cache` with CacheName Enum values.

hub/policy/src/main/java/uk/gov/ida/hub/policy/factories/SessionStoreFactory.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private SessionStoreFactory() {}
+
+    public static TransitionStore<State> getSessionStateStore(
+            InfinispanCacheManager infinispanCacheManager
+    ) {
+        return getSessionStore(infinispanCacheManager, "state_cache");
+    }
+
+    public static ConcurrentMap<SessionId, DateTime> getSessionExpirationStore(
```
adityapahuja: getSessionExpirationStore() method is not used anywhere in policy. Is the method needed in the future?

hub/policy/src/main/java/uk/gov/ida/hub/policy/PolicyApplication.java
```diff
@@ -66,11 +67,19 @@ public final void initialize(Bootstrap<PolicyConfiguration> bootstrap) {
...
+
+        final SessionBundle sessionBundle = new SessionBundle(infinispanBundle.getInfinispanCacheManagerProvider());
+
+        // the infinispan cache manager needs to be lazy loaded because it is not initialized at this point.
+        bootstrap.addBundle(sessionBundle);
         guiceBundle = GuiceBundle.defaultBuilder(PolicyConfiguration.class)
-                .modules(getPolicyModule(), new EventEmitterModule(),  bindInfinispan(infinispanBundle.getInfinispanCacheManagerProvider()))
+                .modules(getPolicyModule(),
+                        new EventEmitterModule(),
+                        getSessionModule(infinispanBundle.getInfinispanCacheManagerProvider(),
```
adityapahuja: Can we replace this line with `new SessionModule(infinispanBundle.getInfinispanCacheManagerProvider(), sessionBundle.getSessionStateStoreProvider())` and remove `getSessionModule()` method from this class? Less code -> easy to read.

hub/policy/src/main/java/uk/gov/ida/hub/policy/configuration/InfinispanConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Valid
+    @JsonProperty
+    protected boolean enabled;
+
+    @Valid
+    @JsonProperty
+    private uk.gov.ida.shared.dropwizard.infinispan.config.InfinispanConfiguration configuration;
+
+    public boolean isEnabled() {
```
adityapahuja: `isEnabled()` method is used in `StateControllerFactoryTest.java` only. Is the method needed in the future?

adityapahuja:
There are some minor changes. All other changes look good.

andy-paine:
This is probably of a lot lower value if we are going to have to fail over to a new platform anyway. If we still want the concept it is still easy enough to pull out and the `SessionStore` stuff in the Redis PR I just raised is far simpler anyway so it would be quicker to just re-do this.

PR #114, opened 2018-08-16
EID-866: Add extra eidas and Sha1 data to MSA HealthCheck report

FelixMarcusMillne:


FelixMarcusMillne:
Nothing to do with us, but removed them for this class anyway to keep Codacy happy.

PR #131, opened 2018-09-03
Hub 271 move config to hub

rachelthecodesmith:
Add two endpoints:

-get all enabled, single idp enabled idps
-get all enabled, single idp enabled rps


hub/config/src/main/java/uk/gov/ida/hub/config/resources/TransactionsResource.java
```diff
@@ -124,6 +124,22 @@ public MatchingProcessDto getMatchingProcess(
                 .collect(Collectors.toList());
     }
 
+    @GET
+    @Path(Urls.ConfigUrls.SINGLE_IDP_ENABLED_TRANSACTIONS_LIST_PATH)
+    @Timed
+    public List<TransactionDetailedDisplayData> getSingleIDPEnabledServiceListTransactions(){
```
chrisclayson: My temptation would have been to update the `getEnabledServiceListTransactions()` method as we wrote that specifically for this journey.

rachelthecodesmith: Ah, I assumed that was pre-existing rather than created for single idp, sorry should have remembered/checked will make the amendment.

PR #138, opened 2018-09-07
EID-939 - Update latest saml-libs to fix bug

JHjava:
- The bug in saml-libs means that it will not throw an exception
when keyinfo is required and the signing cert is not there. This has been
resolved in the latest version of saml-libs.
- Remove meaningless comment in code
- Add test to ensure exception is thrown when signing cert is not present

### DIFF
--- a/hub-saml/src/test/java/uk/gov/ida/saml/hub/api/HubTransformersFactoryTest.java
+++ b/hub-saml/src/test/java/uk/gov/ida/saml/hub/api/HubTransformersFactoryTest.java
@@ -4,10 +4,10 @@ import org.apache.xml.security.exceptions.Base64DecodingException;
 import org.apache.xml.security.utils.Base64;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.rules.ExpectedException;
 import org.opensaml.saml.saml2.core.AuthnRequest;
-import org.opensaml.saml.saml2.core.Response;
 import org.opensaml.xmlsec.algorithm.DigestAlgorithm;
 import org.opensaml.xmlsec.algorithm.SignatureAlgorithm;
 import org.opensaml.xmlsec.algorithm.descriptors.DigestSHA256;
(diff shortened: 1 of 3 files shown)


hub-saml/src/test/java/uk/gov/ida/saml/hub/api/HubTransformersFactoryTest.java
```diff
@@ -37,6 +36,9 @@
 public class HubTransformersFactoryTest {
     private StringToOpenSamlObjectTransformer<AuthnRequest> stringtoOpenSamlObjectTransformer;
 
+    private final SignatureAlgorithm signatureAlgorithm = new SignatureRSASHA256();
```
IreneLau-GDS: 👍 Good tidy up.

hub-saml/src/test/java/uk/gov/ida/saml/hub/api/HubTransformersFactoryTest.java
```diff
@@ -47,57 +49,59 @@ public void setUp() {
...
         Assert.assertNotNull(apply);
 
-        //And the Authn Request contains a KeyInfo section for eIDAS
         AuthnRequest authnReq = stringtoOpenSamlObjectTransformer.apply(apply);
         Assert.assertNotNull(authnReq);
         Assert.assertNotNull("The Authn Request contains a KeyInfo section for eIDAS", authnReq.getSignature().getKeyInfo());
     }
 
-    private static IdaKeyStore getKeyStore() throws Base64DecodingException {
+    @Test (expected = SamlTransformationErrorException.class)
```
IreneLau-GDS: Is it worth capturing and asserting the text of the exception? This is because there are multiple places can throw the same exception for different reasons... 🤔

JHjava: Yes, that's a good point. I'll make the change.

hub-saml/src/test/java/uk/gov/ida/saml/hub/api/HubTransformersFactoryTest.java
```diff
@@ -47,57 +53,61 @@ public void setUp() {
...
+                digestAlgorithm
+        );
+        EidasAuthnRequestFromHub eidasAuthnRequestFromHub = EidasAuthnRequestBuilder.anEidasAuthnRequest()
+                .withLevelsOfAssurance(Collections.singletonList(AuthnContext.LEVEL_2))
+                .buildFromHub();
+
+        eidasTransformer.apply(eidasAuthnRequestFromHub);
+    }
+
+    private static IdaKeyStore getKeyStore(boolean includeSigningCert) throws Base64DecodingException {
```
vixus0: Bit opaque having the signing cert inclusion hidden in a boolean parameter but should do for now.

PR #142, opened 2018-09-13
BAU: Let puppet set memory allocations

andy-paine:
- I mistakenly made all the memory allocations for the apps the same
- This puts the allocations into an env var so that puppet can set it as
  it used to when it layed down the upstart files


stephengrier:
Related to this: https://github.com/alphagov/verify-puppet/pull/1330

PR #143, opened 2018-09-17
HUB-264 - Add Headless Redirect Links to Pause Page

chrisclayson:
Add new config value to transaction resources for the service's headless journey start page.
Ensure new value is returned in enabled transaction list

### DIFF
--- a/hub/config/src/main/java/uk/gov/ida/hub/config/resources/TransactionsResource.java
+++ b/hub/config/src/main/java/uk/gov/ida/hub/config/resources/TransactionsResource.java
@@ -80,7 +80,8 @@ public class TransactionsResource {
         return new TransactionDisplayData(
                 configData.getSimpleId().orElse(null),
                 configData.getServiceHomepage(),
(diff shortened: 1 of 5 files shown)


hub/config/src/main/java/uk/gov/ida/hub/config/resources/TransactionsResource.java
```diff
@@ -80,7 +80,8 @@ private TransactionDisplayData buildTransactionDisplayData(TransactionConfigEnti
         return new TransactionDisplayData(
                 configData.getSimpleId().orElse(null),
                 configData.getServiceHomepage(),
-                configData.getLevelsOfAssurance());
+                configData.getLevelsOfAssurance(),
+                configData.getHeadlessStartpage());
```
jakubmiarka: what happens if we don't have the headless start page in config?

chrisclayson: It will be `null`.

PR #145, opened 2018-09-18
BAU: Update to use ObjectMapper provided by DropWizard

adityapahuja:
Modules classes create a new instance of ObjectMapper that is not same as the instance of ObjectMapper provided by DropWizard. This results in logging timestamp not following the ISO8601DateFormat. The fix is to get Modules classes to use the existing ObjectMapper provided by the DropWizard instead of a new instance of ObjectMapper. This will allow applications to log timestamp in the ISO8601DateFormat.

I ran pre-commit and hub acceptance test. They passed.

Authors: @adityapahuja

### DIFF
--- a/hub/config/src/main/java/uk/gov/ida/hub/config/ConfigModule.java
+++ b/hub/config/src/main/java/uk/gov/ida/hub/config/ConfigModule.java
@@ -1,12 +1,12 @@
 package uk.gov.ida.hub.config;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.datatype.guava.GuavaModule;
 import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
 import com.google.inject.TypeLiteral;
 import io.dropwizard.configuration.ConfigurationFactoryFactory;
 import io.dropwizard.configuration.DefaultConfigurationFactoryFactory;
+import io.dropwizard.setup.Environment;
 import uk.gov.ida.common.shared.security.X509CertificateFactory;
 import uk.gov.ida.common.shared.security.verification.CertificateChainValidator;
 import uk.gov.ida.common.shared.security.verification.OCSPCertificateChainValidator;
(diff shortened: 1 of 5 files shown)


hub/config/src/main/java/uk/gov/ida/hub/config/ConfigModule.java
```diff
@@ -73,7 +74,6 @@ protected void configure() {
         bind(new TypeLiteral<ConfigEntityDataRepository<CountriesConfigEntityData>>(){}).asEagerSingleton();
         bind(new TypeLiteral<ConfigEntityDataRepository<MatchingServiceConfigEntityData>>(){}).asEagerSingleton();
         bind(new TypeLiteral<ConfigEntityDataRepository<IdentityProviderConfigEntityData>>(){}).asEagerSingleton();
-        bind(ObjectMapper.class).toInstance(new ObjectMapper().registerModule(new GuavaModule()));
```
vixus0: D'you know what the `new GuavaModule()` was for?

adityapahuja: I think GuavaModule is used for deserialising and serialising Optional.absent().

This line creates a new instance of ObjectMapper and it does not have GuavaModule registered. This instance is not from DropWizard. 

This pull request changes Config application to use the existing instance of ObjectMapper provided by DropWizard (created in bootstrap) and this instance already has GuavaModule registered. Please see https://github.com/dropwizard/dropwizard/blob/fb58944deddfa2925adb2fd4a3b4c3dba5998cd7/dropwizard-jackson/src/main/java/io/dropwizard/jackson/Jackson.java#L55.

adityapahuja: `This line creates a new instance of ObjectMapper and it does not have GuavaModule registered. This instance is not from DropWizard.` should be read as `This line creates a new instance of ObjectMapper and it does not have GuavaModule registered (i.e. it has to manually register GuavaModule) in order to make it work. This instance is not from Dropwizard.`

vixus0: Ok, sounds good to me.

adityapahuja: 👍

PR #154, opened 2018-10-02
Cert from metadata

kerrr:
Created `SigningCertFromMetadataExtractor` which is used by an overridden `IdaKeyStoreCredentialRetriever` to retrieve Certificates from metadata.

Bind these new objects in the `SamlEngineModule` and remove old binding for `IdaKeyStoreCredentialRetriever`

Add tests for `SigningCertFromMetadataExtractor`

Fix two existing tests that failed when run in isolation in Travis by adding `@RunWith(OpenSAMLMockitoRunner.class)`


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
```diff
@@ -446,21 +447,20 @@ private JsonClient jsonClient(JsonResponseProcessor jsonResponseProcessor, Envir
...
             PrivateKey secondaryEncryptionKey = privateKeyStore.get(KeyPosition.SECONDARY);
             PrivateKey signingKey = privateSigningKey(configuration);
+            PublicKey publicSigningKey = KeySupport.derivePublicKey(signingKey);
 
             KeyPair primaryEncryptionKeyPair = new KeyPair(KeySupport.derivePublicKey(primaryEncryptionKey), primaryEncryptionKey);
             KeyPair secondaryEncryptionKeyPair = new KeyPair(KeySupport.derivePublicKey(secondaryEncryptionKey), secondaryEncryptionKey);
-            KeyPair signingKeyPair = new KeyPair(KeySupport.derivePublicKey(signingKey), signingKey);
+            KeyPair signingKeyPair = new KeyPair(publicSigningKey, signingKey);
+
+            X509Certificate signingCertificate = signingCertExtractor.getSigningCertForCurrentSigningKey(publicSigningKey);
```
rhowe-gds: Should we throw something here if this fails to find a cert?

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+            new EntityIdCriterion(hubEntityId),
+            new UsageCriterion(UsageType.SIGNING),
+            new ProtocolCriterion(SAMLConstants.SAML20P_NS),
+            new EntityRoleCriterion(SPSSODescriptor.DEFAULT_ELEMENT_NAME),
+            new EvaluablePublicKeyCredentialCriterion(publicSigningKey)
+        );
+        try {
+            Credential credential = credentialResolver.resolveSingle(criteriaSet);
+            if (credential instanceof X509Credential) {
+                return ((X509Credential)credential).getEntityCertificate();
```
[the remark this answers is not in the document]
kerrr: Good call, I'll do that.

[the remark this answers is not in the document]
willp-bl: That's a good addition.  It would be nice to test the CriteriaSet in some way, but perhaps that isn't possible.  Are there eidas tests that would fail if the correct signing cert was not found in the metadata?

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
```diff
@@ -48,7 +48,10 @@ public X509Certificate getSigningCertForCurrentSigningKey(PublicKey publicSignin
         try {
             Credential credential = credentialResolver.resolveSingle(criteriaSet);
             if (credential instanceof X509Credential) {
-                return ((X509Credential)credential).getEntityCertificate();
+                X509Certificate x509Cert = ((X509Credential)credential).getEntityCertificate();
+                if (x509Cert.getPublicKey().equals(publicSigningKey)){
```
willp-bl: minor style thing that you can feel free to ignore - might be better to swap the check around as publicSigningKey will never be null

rhowe-gds:
@richardTowers any chance of a re-review?

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractorTest.java
```diff
@@ -0,0 +1,103 @@
...
+                .addSpServiceDescriptor(SPSSODescriptorBuilder.anSpServiceDescriptor()
+                        .addKeyDescriptor(secondKeyDescriptor)
+                        .build())
+                .build();
+    }
+
+    private X509Certificate hubPrimarySigningCert = new X509CertificateFactory().createCertificate(TestCertificateStrings.HUB_TEST_PUBLIC_SIGNING_CERT);
+    private X509Certificate hubSecondarySigningCert = new X509CertificateFactory().createCertificate(TestCertificateStrings.HUB_TEST_SECONDARY_PUBLIC_SIGNING_CERT);
+
+    //Intentionally deriving these from a different object instance.
```
richardTowers: I don't understand this comment

kerrr: I wanted to make sure the public keys we make the call with are not the same object instances that are in the certificates that we supply in the metadata. In the extractor we check that the public key in the cert we get back is the same as the public key we request with, but outside a test like this they will not be the same object instance. So making that the case here makes the test more lifelike. I wanted the comment to highlight that we were doing this intentionally as otherwise if you need the public key you can just get it out of the cert we create a couple of lines above rather than creating the cert object again.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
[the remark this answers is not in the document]
rhowe-gds: Stub country has a runtime check that the signing cert is in the Authn request KeyInfo and we have code in saml engine which checks that the public key in the certificate is the same as the public key derived from the private key here: https://github.com/alphagov/verify-hub/pull/154/files#diff-88e1cfadc6ee3e788caa52b0b8836223R52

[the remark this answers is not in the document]
rhowe-gds: However, production is the only environment which has multiple signing certificates in the metadata

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
```diff
@@ -0,0 +1,62 @@
...
+                    return x509Cert;
+                }
+            }
+
+        } catch (ResolverException e) {
+            throw new UnableToResolveSigningCertsForHubException(e);
+        }
+        throw new CertificateForCurrentPrivateSigningKeyNotFoundInMetadataException();
```
IreneLau-GDS: This exception class has a super long name... Also, do we really need two different types of exception? 

If the purpose of this method is to "extract public cert", we can potentially have one type of exception and make use of the message field to describe what kind of error it is.

Something like `SigningKeyExtractionException` and we do:
```
    throw new SignKeyExtractionException("Certificate for current private signing key not found in metadata.");
```

Then for line 58...
```
    throw new SignKeyExtractionException("Unable to resolve signing cert for hub.");
```

hub-saml/src/test/java/uk/gov/ida/saml/metadata/HubMetadataPublicKeyStoreTest.java
```diff
@@ -25,6 +27,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.fail;
 
+@RunWith(OpenSAMLMockitoRunner.class)
```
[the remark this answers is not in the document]
rhowe-gds: Some tests were failing when run in isolation
Also they were failing in Travis as well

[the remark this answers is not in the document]
kerrr: Travis was failing running those tests with an NPE. This was strange as running all the tests locally was all green. Narrowed it down to the fact the tests failed if run in isolation. Narrowed it down to find that running another test that had `@RunWith(OpenSAMLMockitoRunner.class)` before hand allowed the failing tests to complete successfully. Adding that into the failing tests allowed them to run in isolation correctly. Didn't dig to deep into what was going on.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
kerrr: Yes, this is probably a good idea. Just went with what we had from Will P on this one.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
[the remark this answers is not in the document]
willp-bl: That's why I think it should be tested - if it goes wrong the place it is most likely to be discovered is prod (sadface)

There was [previously] no test in this repo that checked the outbound message had the correct cert in KeyInfo, and that would be a good proxy to check that this code is working...  

This PR exists because the cert was removed then had to be re-added after the issue was discovered in a prod-like environment - it would be good to know that this has broken as early as possible

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractorTest.java
richardTowers: That makes sense. I think the comment could be clearer though - perhaps `Intentionally deriving these PublicKeys from new Certificate instances so they don't have the same object identity as elsewhere` ?

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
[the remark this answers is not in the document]
willp-bl: The CriteriaSet is the thing that sifts the right bit out of the metadata, and that is not tested right now, but ultimately that might not be very easy so :shrug:

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
```diff
@@ -0,0 +1,60 @@
...
+        try {
+            Credential credential = credentialResolver.resolveSingle(criteriaSet);
+            if (credential instanceof X509Credential) {
+                X509Certificate x509Cert = ((X509Credential)credential).getEntityCertificate();
+                if (x509Cert.getPublicKey().equals(publicSigningKey)){
+                    return x509Cert;
+                }
+            }
+        } catch (ResolverException e) {
+            throw new SigningKeyExtractionException(e);
```
IreneLau-GDS: Don't you want to put a message in it?

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/metadata/SigningCertFromMetadataExtractor.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+        throw new SigningKeyExtractionException();
```
IreneLau-GDS: I think you should put a message in it.

IreneLau-GDS: See my original comment. You now have two error cases for one exception type, the idea is to put the reason in the message field to distinguish the problems.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
```diff
@@ -656,10 +656,15 @@ private DecoratedSamlResponseToIdaResponseIssuedByIdpTransformer getResponseToIn
                 hubEntityId);
     }
 
+    @Provides
+    @Singleton
+    private SigningCertFromMetadataExtractor getSigningCertFromMetadataExtractor(@Named(VERIFY_METADATA_RESOLVER) MetadataResolver metadataResolver, @Named("HubEntityId") String hubEntityId) throws ComponentInitializationException {
```
IreneLau-GDS: Just to make you aware of this --> We are currently using the same stuff signing keys to sign both Verify and eIDAS AuthnRequest. 

If this changes in the future (i.e. if we decide that the Connector should sign with a different key), this logic will break.

IreneLau-GDS: Also to add - it is actually a bit confusing. 

It's explicitly looking at the federation's metadata and key to populate the eIDAS's AuthnRequest as a Connector. 

It works because of the current design but logically it doesn't quite makes sense...

PR #155, opened 2018-10-03
EID-1004: Fix transition between Verify and eIDAS journeys (part 1/2)

CharlesIC:
Currently, if a user selects a country or an IDP without completing verification and clicks the back button to the journey picker page and then attempts a different type of journey, they receive a "service unavailable" error page as Policy doesn't allow a transition between `IdpSelected` and `EidasCountrySelected` states. The solution to this is to allow the frontend to ask Policy to restart the journey back to `SessionStartedState` when a user goes back to the journey picker page (`/prove-identity`) and selects a different type of journey. We already allow this for some eIDAS states.

This piece of work makes the interfaces more general so that we can apply the same to the `IdpSelected` state later. This commit only affects eIDAS -> Verify transitions. Transitions in the opposite direction will be enabled in a later release.

Summary of changes:
  - Rename `EidasUnsuccessfulJourney` state and controller interfaces to `RestartJourneyState` to make them more general
  - Make the `EidasCountrySelected` state and controller implement the above interface to give the `EidasCountrySelectedState` the ability to be reset to SessionStartedState
  - Add new more general URL to Policy that frontend can call to restart a journey. The old URL will be removed in a later release after a frontend release
  - Add tests to reflect the above changes

_Note:_ This has to be released before the related frontend changes in alphagov/verify-frontend#589

### DIFF
--- a/hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
+++ b/hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
@@ -190,6 +190,20 @@ public class AuthnRequestFromTransactionResourceIntegrationTest {
     }
 
     @Test
+    public void shouldRestartJourney() {
+        sessionId = SessionId.createNewSessionId();
+        TestSessionResourceHelper.createSessionInEidasAuthnFailedErrorState(sessionId, client, buildUriForTestSession(EIDAS_AUTHN_FAILED_STATE, sessionId));
+
+        URI uri = UriBuilder.fromPath("/policy/received-authn-request" + Urls.PolicyUrls.AUTHN_REQUEST_RESTART_JOURNEY_PATH).build(sessionId);
+        Response response = client.target(policy.uri(uri.toASCIIString())).request().post(null);
+
+        assertThat(response.getStatus()).isEqualTo(Response.Status.NO_CONTENT.getStatusCode());
+
+        Response checkStateChanged = client.target(buildUriForTestSession(GET_SESSION_STATE_NAME, sessionId)).request().get();
+        assertThat(checkStateChanged.readEntity(String.class)).isEqualTo(SessionStartedState.class.getName());
+    }
+
+    @Test
     public void getSignInProcessDto_shouldReturnSignInDetailsDto(){
         SessionId session = SessionId.createNewSessionId();
         TestSessionResourceHelper.createSessionInIdpSelectedState(session, samlResponse.getIssuer(), idpEntityId, client,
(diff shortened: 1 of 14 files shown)


hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
```diff
@@ -189,6 +189,20 @@ public void shouldRestartEidasJourney() {
         assertThat(checkStateChanged.readEntity(String.class)).isEqualTo(SessionStartedState.class.getName());
     }
 
+    @Test
+    public void shouldRestartJourney() {
+        sessionId = SessionId.createNewSessionId();
+        TestSessionResourceHelper.createSessionInEidasAuthnFailedErrorState(sessionId, client, buildUriForTestSession(EIDAS_AUTHN_FAILED_STATE, sessionId));
```
simonwo: Is this the only state we need to test for? From the blurb I thought it was also the `IDPSelected` state.

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/EidasCountrySelectedStateController.java
```diff
@@ -214,4 +214,23 @@ private State createEidasAuthnFailedErrorState() {
...
+        stateTransitionAction.transitionTo(sessionStartedState);
+    }
+
+    private SessionStartedState createSessionStartedState() {
+        return new SessionStartedState(
+                state.getRequestId(),
+                state.getRelayState().orNull(),
+                state.getRequestIssuerEntityId(),
+                state.getAssertionConsumerServiceUri(),
+                null,
```
simonwo: Why is this `null`?

simonwo: Turns out this is actually an issue – we need to be preserving the state of the `ForceAuthentication` flag so that if the user switches from Verify to eIDAS and back again we still pass the flag to the IDP.

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/AuthnRequestFromTransactionResourceIntegrationTest.java
CharlesIC: That's in part 2 of the Hub release which will enable `IdpSelected` -> `SessionStarted` transition

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/EidasCountrySelectedStateController.java
CharlesIC: That's correct - I will raise a separate PR for that

CharlesIC: Resolved in https://github.com/alphagov/verify-hub/pull/161

PR #161, opened 2018-10-19
[DO NOT MERGE] EID-1004: "Service Unavailable" error when switching journeys.

dbes-gds:
Add forceAuthenticationForNonEidasJourney field to Eidas states to hold the forceAuthentication flag.
Modify StateControllers to pass the flag between states in the journey.
Update unit tests and builders to test the new flag.

Co-authored-by: Chris Wynne <christopher.wynne@digital.cabinet-office.gov.uk>
Co-authored-by: dbes-gds <daniel.besbrode@digital.cabinet-office.gov.uk>


hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/state/EidasUserAccountCreationRequestSentState.java
```diff
@@ -1,5 +1,6 @@
 package uk.gov.ida.hub.policy.domain.state;
 
+import com.google.common.base.Optional;
```
simonwo: Where is this used?

simonwo: Where is this used?

simonwo: Where is this used?

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/state/EidasCycle3MatchRequestSentState.java
```diff
@@ -1,5 +1,6 @@
 package uk.gov.ida.hub.policy.domain.state;
 
+import com.google.common.base.Optional;
```
simonwo: Where is this used?

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/state/EidasCountrySelectedState.java
```diff
@@ -18,6 +18,7 @@
     private String countryEntityId;
     private final Optional<String> relayState;
     private List<LevelOfAssurance> levelsOfAssurance;
+    private final Boolean forceAuthenticationForNonEidasJourney;
```
simonwo: So how come we don't implement this on a base class? Couldn't we implement `forceAuthentication` on `AbstractState`? To ask this question another way, how many states now _don't_ have a `forceAuthentication` property?

simonwo: So how come we don't implement this on a base class? Couldn't we implement `forceAuthentication` on `AbstractState`? To ask this question another way, how many states now _don't_ have a `forceAuthentication` property?

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/state/IdpSelectedState.java
```diff
@@ -42,7 +42,15 @@ public IdpSelectedState(
...
 
-        super(requestId, requestIssuerId, sessionExpiryTimestamp, assertionConsumerServiceUri, sessionId, transactionSupportsEidas);
+        super(
+            requestId,
+            requestIssuerId,
+            sessionExpiryTimestamp,
+            assertionConsumerServiceUri,
+            sessionId,
+            transactionSupportsEidas,
+            null
```
simonwo: Why `null` here?

simonwo: This isn't an end state is it?

Wynndow: The null is to maintain the existing behaviour; this state never had `forceAuthentication`, and so we don't introduce it here.

I'm assuming the reason it's not included with this state is that `forceAuthentication` is consumed by the IDP, and that by this point in the journey it's been consumed and is no longer required? That's a guess though...

CharlesIC: I think it's fine to leave it out for now as like you said Chris is not used anywhere at the moment and it also means we don't have to make collateral changes to the tests, etc

PR #163, opened 2018-10-24
EID-1004: Minor clean-up and tidy-up after adding forceAuthentication flag to all states

CharlesIC:
Summary of changes:
  - Move `createSessionStartedState` method from subclasses of `AbstractAuthnFailedErrorStateController` to base class
  - Remove `relayState` from `AuthnFailedErrorState` and `UserAccountCreationFailedState` as the flag was moved to the base class in a previous release
  - Remove `forceAuthentication` from `IdpSelectedState` and `RequesterErrorState` as the flag has been moved to the base class
  - Make the `forceAuthentication` flag non-null by default in test state builders so we don't miss it in deserialisation test
  - Update deserialisation test


hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/state/AbstractAuthnFailedErrorState.java
```diff
@@ -10,6 +10,7 @@
 public abstract class AbstractAuthnFailedErrorState extends AbstractState implements ResponsePreparedState {
 
     private static final long serialVersionUID = 8101005936409595481L;
+
```
Wynndow: Is this a style thing?

hub/policy/src/test/java/uk/gov/ida/hub/policy/domain/state/serialization/DeserializationTest.java
```diff
@@ -102,10 +102,10 @@ public void teardown() {
     }
 
     @Test
-    public void shouldDeserializeSessionStartedState() throws Exception {
```
Wynndow: Why can this be removed?

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/state/AbstractAuthnFailedErrorState.java
CharlesIC: Yes, just to keep it consistent with other state controllers

hub/policy/src/test/java/uk/gov/ida/hub/policy/domain/state/serialization/DeserializationTest.java
CharlesIC: Because this method would never throw a checked exception. Java has an exception propagation mechanism that defines this. It's usually not obvious when a method may or may not throw an unchecked exception - I rely on the IDE (IntelliJ) to tell me

PR #165, opened 2018-11-07
EID-1053: Fixing bug in eIDAS journey when user presses back after authentication fails.

tomrosier:
Added in the handling of a EidasAuthnFailedErrorState to Countries Service,
which should trigger going back to a session started state when landing on the country selection page.


hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/EidasAuthnFailedErrorStateController.java
```diff
@@ -23,4 +23,9 @@ public void transitionToSessionStartedState() {
         hubEventLogger.logSessionMovedToStartStateEvent(sessionStartedState);
         stateTransitionAction.transitionTo(sessionStartedState);
     }
+
+    @Override
+    public void selectCountry(final String countryEntityId) {
+        transitionToSessionStartedState();
```
CharlesIC: This should transition to `EidasCountrySelectedState`

hub/policy/src/main/java/uk/gov/ida/hub/policy/services/CountriesService.java
```diff
@@ -53,8 +54,12 @@ public void setSelectedCountry(SessionId sessionId, String countryCode) {
         Class<? extends State> expectedStateClass = SessionStartedState.class;
         if (sessionRepository.isSessionInState(sessionId, EidasCountrySelectedState.class)) {
             expectedStateClass = EidasCountrySelectedState.class;
+        } else if (sessionRepository.isSessionInState(sessionId, EidasAuthnFailedErrorState.class)) {
+            expectedStateClass = EidasAuthnFailedErrorState.class;
         }
 
+
+
         EidasCountrySelectingStateController eidasCountrySelectingStateController = (EidasCountrySelectingStateController) sessionRepository.getStateController(sessionId, expectedStateClass);
```
CharlesIC: We can completely remove the `expectedStateClass` variable since we always expect the state to implement `EidasCountrySelectingState`

PR #171, opened 2018-11-21
BAU: Add systemd files

adityapahuja:
Since Ubuntu 15.04, systemd has replaced Upstart as the standard boot and service manager. This commit add systemd files to SAML Proxy, SAML SOAP Proxy, Policy, SAML Engine and Config Java applications and these files are used for starting and stopping the applications in Ubuntu 18.04 LTS. Upstart files will remain since they are needed for Ubuntu 14.04.

Author: @adityapahuja


debian/config/config.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+start() {
+  if [ -r /etc/default/verify ]; then
+    . /etc/default/verify
```
rhowe-gds: `/etc/default/verify` is no longer a thing and can be removed, but it won't do any harm so it won't stop me approving this

PR #174, opened 2018-11-27
EID-1117 - Turn DuplicateAssertionValidator into interface

JHjava:
- So it can be used with a different implementation in proxy node

Co-authored-by: Simon Worthington <simon.worthington@digital.cabinet-office.gov.uk>

### DIFF
--- a/hub-saml/src/main/java/uk/gov/ida/saml/core/validators/assertion/DuplicateAssertionValidatorImpl.java
+++ b/hub-saml/src/main/java/uk/gov/ida/saml/core/validators/assertion/DuplicateAssertionValidatorImpl.java
@@ -1,0 +1,46 @@
+package uk.gov.ida.saml.core.validators.assertion;
+
+import com.google.inject.Inject;
+import org.joda.time.DateTime;
+import org.opensaml.saml.saml2.core.Assertion;
+import uk.gov.ida.saml.hub.exception.SamlValidationException;
+
+import java.util.concurrent.ConcurrentMap;
+
+import static uk.gov.ida.saml.core.errors.SamlTransformationErrorFactory.authnStatementAlreadyReceived;
+import static uk.gov.ida.saml.core.errors.SamlTransformationErrorFactory.duplicateMatchingDataset;
+
+public class DuplicateAssertionValidatorImpl implements DuplicateAssertionValidator {
+
+    private final ConcurrentMap<String, DateTime> duplicateIds;
+
+    @Inject
+    public DuplicateAssertionValidatorImpl(ConcurrentMap<String, DateTime> duplicateIds) {
+        this.duplicateIds = duplicateIds;
+    }
+
+    @Override
+    public void validateAuthnStatementAssertion(Assertion assertion) {
+        if (!valid(assertion))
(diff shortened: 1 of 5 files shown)


hub-saml/src/main/java/uk/gov/ida/saml/core/validators/assertion/DuplicateAssertionValidatorImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    private final ConcurrentMap<String, DateTime> duplicateIds;
+
+    @Inject
+    public DuplicateAssertionValidatorImpl(ConcurrentMap<String, DateTime> duplicateIds) {
+        this.duplicateIds = duplicateIds;
+    }
+
+    @Override
+    public void validateAuthnStatementAssertion(Assertion assertion) {
+        if (!valid(assertion)) throw new SamlValidationException(authnStatementAlreadyReceived(assertion.getID()));
```
tomrosier: Drop the throws onto a new line 😄

hub-saml/src/main/java/uk/gov/ida/saml/core/validators/assertion/DuplicateAssertionValidatorImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public void validateMatchingDataSetAssertion(Assertion assertion, String responseIssuerId) {
+        if (!valid(assertion)) throw new SamlValidationException(duplicateMatchingDataset(assertion.getID(), responseIssuerId));
```
tomrosier: Drop throws onto a new line

hub-saml/src/main/java/uk/gov/ida/saml/core/validators/assertion/DuplicateAssertionValidatorImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    private boolean valid(Assertion assertion) {
+        if (isDuplicateNonExpired(assertion)) return false;
```
tomrosier: Drop return onto new line

hub-saml/src/main/java/uk/gov/ida/saml/core/validators/assertion/DuplicateAssertionValidatorImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+        DateTime expire = assertion.getSubject().getSubjectConfirmations().get(0).getSubjectConfirmationData().getNotOnOrAfter();
```
tomrosier: This is a very long line, what happens if any of the elements are null?

JHjava: I believe they are checked before this is called.

Wynndow:
I'm struggling to understand the usefulness of this. I thought it might be because it allowed you to give a different implementation of `duplicateAssertionValidator` to `AuthnStatementAssertionValidator` - which would have made sense to me if `AuthnStatementAssertionValidator` was shared between the hub and the proxy-node - but it isn't. Am I missing something?

PR #176, opened 2018-11-28
lKH7U7yW: Allow lists of assertions in response

alan-gds:
- For the non-matching journey, we need to support sending a List of assertions in ResponseFromHubDto, rather than just one.
- The new List of assertions is used everywhere internally, but to achieve ZDD, we currently retain the encryptedMatchingServiceAssertion field.  Both the list and the original field get populated.

https://trello.com/c/lKH7U7yW/185-wrap-assertions-in-hub-response-to-handle-idp-assertions

Solo: @alan-gds

### DIFF
--- a/hub-saml/src/main/java/uk/gov/ida/saml/hub/transformers/outbound/OutboundResponseFromHubToSamlResponseTransformer.java
+++ b/hub-saml/src/main/java/uk/gov/ida/saml/hub/transformers/outbound/OutboundResponseFromHubToSamlResponseTransformer.java
@@ -1,5 +1,6 @@
 package uk.gov.ida.saml.hub.transformers.outbound;
 
+import org.opensaml.saml.saml2.core.EncryptedAssertion;
 import org.opensaml.saml.saml2.core.Response;
 import org.opensaml.saml.saml2.core.Status;
 import uk.gov.ida.saml.core.OpenSamlXmlObjectFactory;
@@ -7,6 +8,9 @@ import uk.gov.ida.saml.core.domain.OutboundResponseFromHub;
 import uk.gov.ida.saml.core.domain.TransactionIdaStatus;
 import uk.gov.ida.saml.core.transformers.outbound.IdaResponseToSamlResponseTransformer;
 import uk.gov.ida.saml.core.transformers.outbound.IdaStatusMarshaller;
+
+import java.util.List;
+import java.util.stream.Collectors;
 
 public class OutboundResponseFromHubToSamlResponseTransformer extends IdaResponseToSamlResponseTransformer<OutboundResponseFromHub> {
 
@@ -27,9 +31,9 @@ public class OutboundResponseFromHubToSamlResponseTransformer extends IdaRespons
     @Override
     protected void transformAssertions(OutboundResponseFromHub originalResponse, Response transformedResponse) {
         originalResponse
-                .getMatchingServiceAssertion()
+                .getEncryptedAssertions().stream()
                 .map(encryptedAssertionUnmarshaller::transform)
-                .map(transformedResponse.getEncryptedAssertions()::add);
+                .forEach(transformedResponse.getEncryptedAssertions()::add);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


hub-saml/src/main/java/uk/gov/ida/saml/hub/transformers/outbound/OutboundResponseFromHubToSamlResponseTransformer.java
```diff
@@ -27,9 +31,9 @@ public OutboundResponseFromHubToSamlResponseTransformer(
     @Override
     protected void transformAssertions(OutboundResponseFromHub originalResponse, Response transformedResponse) {
         originalResponse
-                .getMatchingServiceAssertion()
+                .getEncryptedAssertions().stream()
                 .map(encryptedAssertionUnmarshaller::transform)
-                .map(transformedResponse.getEncryptedAssertions()::add);
+                .forEach(transformedResponse.getEncryptedAssertions()::add);
```
alan-gds: Not only is forEach() more idiomatic here, it turns out map() won't work if the source stream came from a List rather than an Optional, so this change was crucial.  No idea why that's the case.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/services/RpAuthnResponseGeneratorService.java
```diff
@@ -36,13 +39,16 @@ public AuthnResponseFromHubContainerDto generate(ResponseFromHubDto responseFrom
     private AuthnResponseFromHubContainerDto createSuccessResponse(final ResponseFromHubDto responseFromHub) {
         String authnRequestIssuerEntityId = responseFromHub.getAuthnRequestIssuerEntityId();
 
+        List<String> encryptedAssertions = new ArrayList<>();
+        responseFromHub.getEncryptedMatchingServiceAssertion().ifPresent(encryptedAssertions::add);
```
andy-paine: You could also do `List<String> encryptedAssertions = responseFromHub.getEncrypterMatchingDatasetAssertion().map(Collections::singletonList).orElse(Collections.emptyList())` but given this code is only going to exist for 2 releases there is no point nitpicking

hub-saml/src/main/java/uk/gov/ida/saml/hub/transformers/outbound/OutboundResponseFromHubToSamlResponseTransformer.java
jkunle: I have been learning a little about Streams, don't know much but it could be that it because map is a non-terminal stream operator while foreach is Terminal

`
originalResponse
                .getEncryptedAssertions().stream()
                .map(encryptedAssertionUnmarshaller::transform)
                .map(transformedResponse.getEncryptedAssertions()::add)
                .collect(Collectors.toList());
`
The above also seems to work with map, not sure why map worked with Optional. Perhaps Andy would know.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/services/RpAuthnResponseGeneratorService.java
alan-gds: That's nicer.  I'm always happy to get rid of imperative code.

hub/saml-engine/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlengine/apprule/RpAuthnResponseGeneratorResourceTest.java
```diff
@@ -185,4 +187,102 @@ public void shouldReturnAnErrorResponseGivenBadInput() throws JsonProcessingExce
...
+        Response rpAuthnResponse = client.target(generateAuthnResponseEndpoint).request().post(Entity.entity(responseFromHubDto, MediaType.APPLICATION_JSON_TYPE));
+
+        // Then
+        assertThat(rpAuthnResponse.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+        AuthnResponseFromHubContainerDto result = rpAuthnResponse.readEntity(AuthnResponseFromHubContainerDto
+            .class);
+
+        org.opensaml.saml.saml2.core.Response response = extractResponse(result);
+        assertThat(response.getStatus().getStatusCode().getValue()).isEqualTo(StatusCode.SUCCESS);
+        assertThat(response.getEncryptedAssertions().size()).isEqualTo(2);
```
alan-gds: I would like to do better than just checking the size, but I'm not sure how to mock the encryption etc.

PR #179, opened 2018-11-30
Make Policy send lists of assertions

andy-paine:
- This PR makes policy use the new "encryptedAssertions" field
in ResponseFromHub
- Since we can now do this, also enable it for the non matching journey


andy-paine:
I need this for tomorrow so going to address Alan's comments on another card

andy-paine:
I need this for tomorrow so going to Address Alan's comments in another card

PR #183, opened 2018-12-10
BAU: Fix an NPE

alan-gds:
This issue is blocking hub releases.

I think this change should fix it, but I am having difficulty getting the acceptance tests to run on my machine; it would be really good if a reviewer could do so.

### DIFF
--- a/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/services/RpAuthnResponseGeneratorService.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/services/RpAuthnResponseGeneratorService.java
@@ -41,7 +41,7 @@ public class RpAuthnResponseGeneratorService {
         String authnRequestIssuerEntityId = responseFromHub.getAuthnRequestIssuerEntityId();
 
         List<String> encryptedAssertions = responseFromHub.getEncryptedAssertions();
-        if (encryptedAssertions.isEmpty()) {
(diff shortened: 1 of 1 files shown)


tlwr:
xpost from slack

> surely there will be other codepaths which call `getEncryptedAssertions` which will possibly have this error
> 
> rather than do null checking in many places is it possible to coerce to empty list inside `getEncryptedAssertions`
> verify-hub isn’t my domain so i may not have full context

alan-gds:
Yes, that would be sensible.  It's coming from a DTO class, so the "right" way to sort this out would be to change the Jackson deserialisation settings.  I'm not very familiar with Jackson, and I don't have effective local test coverage, so I'd rather not make that change myself.

Another alternative would be to do the coercion in the DTO manually, but that's not very idiomatic.

Either way, I'm in favour of committing my hack now and finessing it later.

andy-paine:
If you were to merge this an release it you would also be releasing the changes to policy to make it use the list so it would never be null or empty which make this check kind of redundant. The next release is non Zdd anyway unless you revert the policy changes so you might as well just scrap this and say that the fix is just to roll forwards to include the policy changes. This NPE is only present during the transition from matchingAssertion - > list so just finish that transition to fix it

alan-gds:
Right, from what Andy's told me, accepting null encryptedAssertions lists will be actively harmful once the non-ZDD change he mentioned has been released - and also, the non-ZDD change will mean the list will not be null anyway.

So, this PR must NOT be merged; I'm closing it.

PR #187, opened 2018-12-14
QiX0vM8J: Allow non-matching for eIDAS

andy-paine:
- This allows RPs who are using the non matching setting to also consume
  eIDs
- Checks if RP is doing matching and if so returns
  NON_MATCHING_JOURNEY_SUCCESS and moves to relevant state
- Mirrors a prior change in the IdpSelectedState controller


andy-paine:
[eIDAS changes to the hub](https://trello.com/c/QiX0vM8J/241-eidas-changes-to-the-hub)

CharlesIC:
Do we have any transactions configured to use a non-matching journey for acceptance testing?

PR #190, opened 2018-12-17
HUB-399: Create a SIDP customised start page for SA in the hub

chrisclayson:
Added singleIdpStartPageContent attribute to translation data.

Co-authored-by: Kerr Rainey <kerr.rainey@digital.cabinet-office.gov.uk>


hub/config/src/main/java/uk/gov/ida/hub/config/domain/TranslationData.java
```diff
@@ -142,5 +150,15 @@ public Translation setTaxonName(String taxonName) {
             this.taxonName = taxonName;
             return this;
         }
+
+        public Translation setSingleIdpStartPageContent(String singleIdpStartPageContent) {
```
jakubmiarka: Why this needs the `set` methods whereas the custom failed page doesn't?

PR #194, opened 2018-12-20
[demo] consume signed VSP & MSA metadata instead of certs in fedconfig

willp-bl:
when use msa metadata boolean flag is set in fedconfig then the signed
metadata will be retreived from the msa, validated and keys returned
from that

this will use the entityId as the metadata location

this has been tested locally, along with the msa signed metadata change

only one test (twice) that needs to be fixed now

the release of this code as is will not be zero downtime due to the
dto changes


willp-bl:
This is essentially the same code added twice, once to saml-engine and once to saml-soap-proxy

PR #199, opened 2019-01-08
make eventSinkUri an optional configuration parameter

philandstuff:
This removes @NotNull from eventSinkUri in policy, saml proxy and saml
soap proxy.  It also changes their respective modules to register a
no-op EventSinkProxy implementation if eventSinkUri is not provided.

All environments now support the new eventEmitterConfiguration option
instead so eventSinkUri is no longer required.


tlwr:
Green in travis

tlwr:
In old environments event sink still needs to be running

However in new environments (running in parallel to the old environments) we do not want to run event sink / hub support, or a stub event sink.

Phil's change allows us to turn it off, but it will still be turned on where it needs to be

PR #202, opened 2019-01-11
AUTO-102: Use Redis as a session store

andy-paine:
- As part of the move to ECS, we need a session store that does not rely
  on all the nodes having to know about each other
- AWS offer a managed service called Elasticache that we will probably use
  and the proxy node is also using Redis in Kubernetes

Hopefully the commits are fairly self contained but the general tactic for
this PR is:
- Clear up unnecessary tests/code that would get in the way
- Add JSON serialization to all states
- Add abstraction over infinispan that will allow for future
redis implementation of session storage
- Bit more cleanup of configuration class locations
- Use `redisson` library to implement session store against Redis


hub/policy/src/main/java/uk/gov/ida/hub/policy/configuration/SessionStoreConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import uk.gov.ida.shared.dropwizard.infinispan.config.InfinispanConfiguration;
+
+import java.util.Optional;
+
+import static com.google.common.base.Optional.absent;
+
+public class SessionStoreConfiguration {
+
+    @JsonProperty
+    private InfinispanConfiguration infinispan = new InfinispanConfiguration(
```
tomrosier: Can we tidy this up a bit?

hub/policy/src/main/java/uk/gov/ida/hub/policy/SessionStoreConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Optional;
+
+import static com.google.common.base.Optional.absent;
+
+public class SessionStoreConfiguration {
+
+    @JsonProperty
+    private InfinispanConfiguration infinispan = new InfinispanConfiguration(
+            absent(), -1, absent(), absent(), CacheType.standalone, com.google.common.base.Optional.of(Duration.hours(2)),
```
tomrosier: Can we tidy this up a bit?

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/support/PolicyAppRuleWithRedis.java
```diff
@@ -0,0 +1,78 @@
...
+                .add(configOverrides)
+                .build();
+        return mergedConfigOverrides.toArray(new ConfigOverride[mergedConfigOverrides.size()]);
+    }
+
+    @Override
+    protected void before() {
+        clientTrustStore.create();
+        try {
+            redis = new RedisServer(6379);
```
tomrosier: Whats the magic number?

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/support/PolicyAppRuleWithRedis.java
```diff
@@ -0,0 +1,78 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        redis.start();
```
tomrosier: What if redis is null?

andy-paine:
@tomrosier has some valid points but given the lifetime of the infinispan hacks is on the order of weeks I'm going to merge this now as-is and just be even happier when I purge all the infinispan stuff 🎉

PR #203, opened 2019-01-14
BAU: Upgrade gradle version plugins to the latest version

adityapahuja:
The latest version of gradle version plugins supports parallel dependency resolution which will reduce the time it takes for Gradle to complete dependency resolution.

Author: @adityapahuja


adityapahuja:
Gradle using the old version of gradle version plugin (v0.17.0) took over 55 minutes to complete dependency updates task.
Gradle using the latest version of gradle version plugin (v0.20.0) took 12 minutes and 10 seconds to complete dependency updates task because of parallel dependency resolution.

PR #205, opened 2019-01-17
EID-1072 Bump saml-libs to version 169

Wynndow:
It pulls in a few new things.

An update to EidasMetadataResolverRepository to throw an exception
instead of an error if it encounters certificate that's expired. This is
useful in saml-engine and saml-proxy.

MetadataResolverBundle has been updated to generate a dropwizard
healthcheck for the metadata, rather than the app itself having to do
it. This meant the healtcheck being created in saml-engine could be
removed.

It allows optional metadata bundling, which required an update to the
config class for saml-engine as well as saml-soap-proxy to support
Optionals being returned by `getMetadataConfiguration`. The metadata is
always not null in these cases to impact is minimal.

### DIFF
--- a/hub/saml-engine/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlengine/apprule/support/SamlEngineAppRule.java
+++ b/hub/saml-engine/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlengine/apprule/support/SamlEngineAppRule.java
@@ -59,7 +59,7 @@ import static uk.gov.ida.saml.core.test.TestEntityIds.HUB_ENTITY_ID;
 import static uk.gov.ida.saml.metadata.ResourceEncoder.entityIdAsResource;
 
 public class SamlEngineAppRule extends DropwizardAppRule<SamlEngineConfiguration> {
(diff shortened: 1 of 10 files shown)


hub/saml-engine/src/integration-test/java/uk/gov/ida/integrationtest/hub/samlengine/apprule/support/SamlEngineAppRule.java
```diff
@@ -59,7 +59,7 @@
 import static uk.gov.ida.saml.metadata.ResourceEncoder.entityIdAsResource;
 
 public class SamlEngineAppRule extends DropwizardAppRule<SamlEngineConfiguration> {
-    private static final String VERIFY_METADATA_PATH = "/uk/gov/ida/saml/metadata/federation";
+    public static final String VERIFY_METADATA_PATH = "/uk/gov/ida/saml/metadata/federation";
```
tomrosier: This isn't a great way of doing it, but its consistent so lets keep it for now. It would probably be worth moving the config out to one place, rather than it living at the top of a class that delivers functionality.

PR #208, opened 2019-01-21
AUTO-125: Add RP Certificates to Prometheus Metrics in Config

adityapahuja:
This pull request updates Config application to send a metric for each relying party's certificate such as:
    
`verify_config_certificate_expiry{entity_id="https://www.stub rp.gov.uk",use="SIGNING",fingerprint="0AF1B0DE..."}`
    
with a value set to the timestamp of the NotAfter date of the certificate in milliseconds.

Author: @adityapahuja


hub/config/src/main/java/uk/gov/ida/hub/config/ConfigModule.java
```diff
@@ -91,6 +93,20 @@ protected void configure() {
         bind(MatchingServiceAdapterService.class);
     }
 
+    @Provides
+    @Singleton
+    private Gauge getGauge() {
+        return Gauge.build("verify_config_certificate_expiry", "Timestamp of NotAfter value of X.509 certificate")
+                    .labelNames("entity_id","use","fingerprint")
+                    .register();
```
philandstuff: I'd lean slightly towards creating and registering this in `PrometheusClientService` (and relying on it being a singleton) unless you really really need to inject it in tests?

Injecting a `Gauge` isn't terribly scalable because once you add a second gauge to your DI container you have to start doing funny things with `@Named`.  I see Gauge as being like Logger in this respect - you don't inject loggers, you new them up in the class that uses them.

adityapahuja: Done.

philandstuff:
To add some context: this PR adds the metrics so that we can alert on them in Prometheus when we switch to the new infrastructure.

Currently, the sensu checks work by hitting Config's `/config/certificates/health-check` endpoint and alerting if any certificates are returned. Config's job is to go through all the expiry dates and compare them against the current time.

In the new world, Config is still the source of truth about RP certificates and expiry dates, but the logic about when to alert will come out of Config and live in our alerting (prometheus/grafana/alertmanager etc).  This PR adds a prometheus metric that reports the expiry dates of certificates, so that we can add alerts in the new infrastructure for upcoming expiry dates.  As a bonus, we can also create dashboards to see further into the future about upcoming required certificate rotations.

phantas:
Fantastic! Are you able to report the CN as well? It's the key identifier that we report to RPs and secops.

adityapahuja:
Yes, I can make a change to add a subject which has CN.

PR #217, opened 2019-01-29
AUTO: Use Redis for SAML Engine replay cache

andy-paine:
- Similar to PR #202 but a little different
- This allows SAML Engine to use redis as the cache/store for the
 assertion IDs and authn IDs it has seen for replay attack detection
- If no redis config is defined, default back to (existing) infinispan
 configuration
- As we don't need to use the `expire` feature of Redis, we can get away
 with using an `RMap` which implements the `ConcurrentMap` interface
- This means we can inject a Redis backed Map instead of Infinispan
  backed map when config is present rather than pulling out an interface
  class as we did in Policy


andy-paine:
Just realised that even though we set our own "expire" date, we still rely on infinispan kicking out old entries to keep the cache size down. I need to do something closer to what happened in Policy to make sure old values are evicted.

PR #221, opened 2019-02-01
AUTO-146: Move last updated from labels to a separate metric

adityapahuja:
Every unique combination of key-value label pairs represents a new time series which can dramatically increase the amount of data stored. We should not use labels to store dimensions with high cardinality (many different label values) such as timestamp or other unbounded sets of values. This commit updates to have pair of metrics (one for expiry date or OCSP revocation status and another one for last updated (timestamp)). The last updated metric is used to indicate whether expiry date metric or OCSP revocation status metric is stale or not. It will be used to alert support team that the metrics are out of date and the team will fix the issue. This commit also adds a serial number to certificate metrics since metadata certificate metrics already have serial numbers.

Author: @adityapahuja

### DIFF
--- a/hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateExpiryDateCheckService.java
+++ b/hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateExpiryDateCheckService.java
@@ -13,33 +13,38 @@ import java.util.Set;
 public class CertificateExpiryDateCheckService implements Runnable {
     private static final Logger LOG = LoggerFactory.getLogger(CertificateExpiryDateCheckService.class);
     private final CertificateService certificateService;
-    private final Gauge gauge;
+    private final Gauge expiryDateGauge;
+    private final Gauge lastUpdatedGauge;
 
     public CertificateExpiryDateCheckService(final CertificateService certificateService,
-                                             final Gauge gauge) {
+                                             final Gauge expiryDateGauge,
+                                             final Gauge lastUpdatedGauge) {
         this.certificateService = certificateService;
-        this.gauge = gauge;
+        this.expiryDateGauge = expiryDateGauge;
+        this.lastUpdatedGauge = lastUpdatedGauge;
     }
 
     @Override
     public void run() {
-        final Set<CertificateDetails> certificateDetailsSet = certificateService.getAllCertificatesDetails();
-        final String TIMESTAMP = DateTime.now(DateTimeZone.UTC).toString();
-        gauge.clear();
-        certificateDetailsSet.forEach(
-            certificateDetails -> {
+        try {
+            final Set<CertificateDetails> certificateDetailsSet = certificateService.getAllCertificatesDetails();
+            final double timestamp = DateTime.now(DateTimeZone.UTC).getMillis();
+            certificateDetailsSet.forEach(certificateDetails -> {
                 try {
-                    gauge.labels(
-                        certificateDetails.getIssuerId(),
+                    expiryDateGauge.labels(certificateDetails.getIssuerId(),
                         certificateDetails.getCertificate().getCertificateType().toString(),
                         certificateDetails.getCertificate().getSubject(),
                         certificateDetails.getCertificate().getFingerprint(),
-                        TIMESTAMP)
+                        String.valueOf(certificateDetails.getCertificate().getSerialNumber()))
                          .set(certificateDetails.getCertificate().getNotAfter().getTime());
                 } catch (CertificateException e) {
                     LOG.warn(String.format("Invalid X.509 certificate [issuer id: %s]", certificateDetails.getIssuerId()));
                 }
             });
-        LOG.info("Updated Certificates Expiry Dates Metrics.");
+            lastUpdatedGauge.set(timestamp);
+            LOG.info("Updated Certificates Expiry Dates Metrics.");
+        } catch (Exception e) {
+            LOG.warn(e.getMessage());
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateExpiryDateCheckService.java
```diff
@@ -13,33 +13,46 @@
...
+    private final Gauge lastUpdatedGauge;
 
     public CertificateExpiryDateCheckService(final CertificateService certificateService,
-                                             final Gauge gauge) {
+                                             final Gauge expiryDateGauge,
+                                             final Gauge lastUpdatedGauge) {
         this.certificateService = certificateService;
-        this.gauge = gauge;
+        this.expiryDateGauge = expiryDateGauge;
+        this.lastUpdatedGauge = lastUpdatedGauge;
```
philandstuff: I don't think we need a lastUpdatedGauge for the expiry date metric. It's important for OCSP because we want to know the last time OCSP was successful, but the expiry date of a certificate isn't something that should ever change over time so we don't need to keep track of updates here.

hub/config/src/main/java/uk/gov/ida/hub/config/application/PrometheusClientService.java
```diff
@@ -12,7 +12,13 @@
 
 public class PrometheusClientService {
     public static final String VERIFY_CONFIG_CERTIFICATE_EXPIRY = "verify_config_certificate_expiry";
+    public static final String VERIFY_CONFIG_CERTIFICATE_EXPIRY_TIMESTAMP = "verify_config_certificate_expiry_timestamp";
     public static final String VERIFY_CONFIG_CERTIFICATE_OCSP_SUCCESS = "verify_config_certificate_ocsp_success";
+    public static final String VERIFY_CONFIG_CERTIFICATE_OCSP_SUCCESS_TIMESTAMP = "verify_config_certificate_ocsp_success_timestamp";
```
philandstuff: can this be `verify_config_certificate_ocsp_last_success_timestamp`? otherwise it's not clear what the timestamp means.

hub/config/src/main/java/uk/gov/ida/hub/config/application/PrometheusClientService.java
```diff
@@ -12,7 +12,13 @@
...
+    public static final String VERIFY_CONFIG_CERTIFICATE_EXPIRY_HELP = "Timestamp Of NotAfter Value Of X.509 Certificate";
+    public static final String VERIFY_CONFIG_CERTIFICATE_EXPIRY_TIMESTAMP_HELP = "Certificate Expiry Date Metric Timestamp";
+    public static final String VERIFY_CONFIG_CERTIFICATE_OCSP_SUCCESS_HELP = "X.509 Certificate OCSP Revocation Status";
+    public static final String VERIFY_CONFIG_CERTIFICATE_OCSP_SUCCESS_TIMESTAMP_HELP = "X.509 Certificate OCSP Revocation Status Metric Timestamp";
```
philandstuff: similarly, can the help text say something like "The timestamp of the last time we got a successful OCSP response"

hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateExpiryDateCheckService.java
adityapahuja: I understand from Andy that Verify team is planning to make changes to Config application to have a dynamic list of RP and MSA certificates during runtime (i.e. it will reload a list of RP and MSA certificates from S3 bucket or other sources every N seconds) rather than loading a list of certificates once at the beginning of the program in the near future. This is the reason why I added the timestamp to the metric.

hub/config/src/main/java/uk/gov/ida/hub/config/application/PrometheusClientService.java
adityapahuja: I have renamed it to `verify_config_certificate_ocsp_revocation_status_last_updated`.

hub/config/src/main/java/uk/gov/ida/hub/config/application/PrometheusClientService.java
adityapahuja: I have updated it to `X.509 Certificate OCSP Revocation Status Metric Last Updated (ms)`.

hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateExpiryDateCheckService.java
philandstuff: if that's the case, I think we only need a single last-updated gauge (for the most recent time we updated), rather than a separate gauge per certificate.  (in other words, I think we can remove all the certificate-type labels from lastUpdatedGauge)

hub/config/src/main/java/uk/gov/ida/hub/config/application/PrometheusClientService.java
philandstuff: `verify_config_certificate_ocsp_revocation_status_last_updated` suggests it will update for any OCSP response, not just successful ones; but I think it's more useful for alerting that we have a metric which only updates when we have a successful OCSP response?

hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateExpiryDateCheckService.java
adityapahuja: Done.

hub/config/src/main/java/uk/gov/ida/hub/config/application/PrometheusClientService.java
philandstuff: We spoke face to face and I linked to the [prometheus best practices for alerting on batch jobs](https://prometheus.io/docs/practices/alerting/#batch-jobs) which feels very similar to this use case. We've agreed to go with `verify_config_certificate_ocsp_last_success_timestamp` 👍

PR #227, opened 2019-02-11
AUTO: Fix entrypoint/cmd in Dockerfile

andy-paine:
- Best practice is to use the exec syntax and define entrypoint as the
  main program and CMD as the command you expect to use
- This will come in handy when we want to do things such as
  fed-config-check with the image


andy-paine:
This isn't as simple as this and I don't have time to think about it right now :(

PR #228, opened 2019-02-19
dGJgfILT: Change response status on duplicate request exception to 400

bjgill:
Stop treating it as an 'internal server error'. Makes no difference to the front-end, which still displays the same 'something has gone wrong' error to the user.

Co-authored-by: John Watts <john.watts@digital.cabinet-office.gov.uk>

### DIFF
--- a/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/exceptions/SamlProxyDuplicateRequestExceptionMapper.java
+++ b/hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/exceptions/SamlProxyDuplicateRequestExceptionMapper.java
@@ -12,7 +12,6 @@ import uk.gov.ida.shared.utils.logging.LevelLoggerFactory;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.core.Response;
-import java.util.Optional;
 import java.util.UUID;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static uk.gov.ida.common.ExceptionType.INVALID_SAML_DUPLICATE_REQUEST_ID;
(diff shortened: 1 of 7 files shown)


bjgill:
There's a possible question as to what error code we should be using. @phantas - do you have a view?

This PR uses 400 (Bad Request). Might 403 (Forbidden) be more precise?

RacingHippo:
I reckon 400 is more descriptively accurate, but.... do we know what causes it?

bjgill:
We're not entirely sure what's causing it. The user is somehow making duplicate requests. We've been unable to repro.

andy-paine:
This exception is no longer sent to the audit system - are we sure that's a change in behaviour we are happy with?

tomrosier:
I think this should be a 400 or a 409. As Andy says we should keep the exception but maybe turn this down from a Error to a Warn

bjgill:
Thanks for the comments. I've added auditing, and actually got the test to check this (it wasn't previously).

Tom - I've used the level defined in the `ExceptionType` - i.e. info-level. Please do shout if you think this should be raised to warn.

Following the comments about the code, I've kept it at 400.

This should then be ready for re-review.

hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/exceptions/SamlProxyDuplicateRequestExceptionMapper.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    protected Response handleException(SamlDuplicateRequestIdException exception) {
+        UUID errorId = UUID.randomUUID();
+
+        Optional<SessionId> sessionId = getSessionId();
+        if (sessionId.isPresent()) {
+            eventSinkMessageSender.audit(exception, errorId, sessionId.get());
+        } else {
+            eventSinkMessageSender.audit(exception, errorId, SessionId.NO_SESSION_CONTEXT_IN_ERROR);
```
tomrosier: `eventSinkMessageSender.audit(` shouldn't be duplicated really if just the last parameter is different.

bjgill: A good point - I checked the API docs for Optional, and learnt about orElse, which does exactly what we need here. I've made the same change to the other 2 incidences of this.

galund:
Will we still be responding to the user with a 500? (Unless we can say the user did something wrong, then presumably so.)

In which case this is fine, as long as we can still effectively investigate that 500. If this puts us in a position where we just have less logging, but the same number of problems as before, then I'm not hugely enthusiastic (even though, in a narrow technical sense, 400 from the hub may well be more correct than 500.)

bjgill:
Yes - the user-facing behaviour is unchanged. We've checked that there is a frontend UT that covers this case, and verifies that the user still sees a `500`.

Likewise, none of the logging has changed. So, I don't think this change will make the error any less diagnosable

andy-paine:
Nice job 👍

PR #229, opened 2019-02-22
AUTO-239: Use lettuce instead of redisson

andy-paine:
- We've been having some issues with Redis in the new environments
- We think this is due to Redisson trying to be clever about the node
 topology which doesn't work for AWS Elasticache
- Lettuce gives us more fine grained control of our connection and tries
 to do less "magic" when talking to Redis
- Redisson also implements it's "Map" as a single key in Redis so we get
 very little interesting metrics about Redis usage
- Use lettuce, similarly to the proxy node so that we have one common
 library within the programme for interacting with Redis
- Done to both policy and saml-engine

The changes required to get this working with lettuce were mostly focussed
around having to do the JSON serialization ourselves, including a custom
codec and a Mixin to get the SessionId to work. Renamed some methods as
well to keep them in line with the field names.


hub/policy/src/main/java/uk/gov/ida/hub/policy/redis/SessionIdMixIn.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.ida.hub.policy.redis;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/* The SessionId class is used for serialization both between microservices
+* and into Redis. To allow it to be serialized in the correct way for Redis,
+* this Mixin is included with the ObjectMapper Redis uses */
```
philandstuff: I'm not sure I understand the point this comment is making.  Are you saying that:

 - SessionId is serialized both between microservices and for redis
 - SessionId is serialized in *different ways* between microservices and for redis
 - therefore, we can't use annotations to specify how to serialize SessionId in all cases
 - this MixIn specifies *just* the redis serialization behaviour

First of all, have I understood you correct?  Second, if so, can you answer some questions?

 - where is the JSON serialization for SessionId between microservices defined?
 - *why* is the serialization different between microservices and for redis?
 - can we make serialization the same so we can just use annotations like we do everywhere else?

andy-paine: > * SessionId is serialized both between microservices and for redis

Yes, it is serialized into DTOs that saml-proxy sends and it is serialized in states that redis stores
> * SessionId is serialized in _different ways_ between microservices and for redis

Yes, Redis uses `@JsonCreator` and ignores getters, the microservices just use pure reflection
> * therefore, we can't use annotations to specify how to serialize SessionId in all cases

Adding the annotations breaks one of the serializations
> * this MixIn specifies _just_ the redis serialization behaviour

Yep

> * where is the JSON serialization for SessionId between microservices defined?

By providing empty constructors, Jackson uses reflection to work out how to populate the fields
> * _why_ is the serialization different between microservices and for redis?

Microservices uses and old and (comparatively) hard to reason about system
> * can we make serialization the same so we can just use annotations like we do everywhere else?

We _could_ but the problem is that the empty constructor/reflection method is used for *all* DTOs so it would require us to annotate every single one, including sub-objects. This is probably beneficial moving forwards but it seems out of scope for this work.

philandstuff: We spoke face to face and we agree that we want SessionId to serialize as a raw string rather than as an object `{"sessionId":"foobarbaz"}`.  Hopefully this will mean we can get rid of the mixin?

hub/policy/src/main/java/uk/gov/ida/hub/policy/redis/SessionStoreRedisCodec.java
```diff
@@ -0,0 +1,56 @@
...
+    public SessionStoreRedisCodec(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public SessionId decodeKey(ByteBuffer keyBytes) {
+        try {
+            InputStream inputStream = new ByteBufferBackedInputStream(keyBytes);
+            // TODO: Replace with `inputStream.readAllBytes()` when we only use Java 9+
+            return new SessionId(new String(ByteStreams.toByteArray(inputStream)));
```
philandstuff:
```suggestion
            return new SessionId(new String(ByteStreams.toByteArray(inputStream), StandardCharsets.UTF_8));
```
this is perhaps the nittiest of picks, because there's almost no situation where the locale can screw up a simple hex string, but I instinctively always specify charset just out of habit

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/redis/AuthnRequestExpirationCacheRedisCodec.java
```diff
@@ -0,0 +1,33 @@
...
+    public AuthnRequestExpirationCacheRedisCodec(ObjectMapper objectMapper) {
+        super(objectMapper);
+    }
+
+    @Override
+    public AuthnRequestIdKey decodeKey(ByteBuffer keyBytes) {
+        try {
+            InputStream inputStream = new ByteBufferBackedInputStream(keyBytes);
+            // TODO: Replace with `inputStream.readAllBytes()` when we only use Java 9+
+            return new AuthnRequestIdKey(new String(ByteStreams.toByteArray(inputStream)));
```
philandstuff:
```suggestion
            return new AuthnRequestIdKey(new String(ByteStreams.toByteArray(inputStream), StandardCharsets.UTF_8));
```

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/redis/AssertionExpirationCacheRedisCodec.java
```diff
@@ -0,0 +1,32 @@
...
+    public AssertionExpirationCacheRedisCodec(ObjectMapper objectMapper) {
+        super(objectMapper);
+    }
+
+    @Override
+    public String decodeKey(ByteBuffer keyBytes) {
+        try {
+            InputStream inputStream = new ByteBufferBackedInputStream(keyBytes);
+            // TODO: Replace with `inputStream.readAllBytes()` when we only use Java 9+
+            return new String(ByteStreams.toByteArray(inputStream));
```
philandstuff:
```suggestion
            return new String(ByteStreams.toByteArray(inputStream), StandardCharsets.UTF_8);
```
oops, missed this one

PR #232, opened 2019-03-04
BAU: Don't error for dual-running encryption certs

vixus0:
Currently, the Hub makes a lot of noise when dual running its old and
new SAML encryption certificates. This shows up in Sentry and floods the
logs. We should only be raising an exception if decryption has failed for all the
certificates.

The main change involves passing a `List<AssertionDecrypter>` to the
`IdpResponseValidator` and `MatchingServiceResponseValidator` who
(despite their name) are also responsible for decrypting the assertions
in the IDP response and MSA response, respectively. The validator tries
to decrypt with each `AssertionDecrypter` in turn, logging a warning on
failure. If the validator can't decrypt with any of the
`AssertionDecrypter`s, it throws a `SamlFailedToDecryptException` which
should appear in our Sentry alerting as usual.


galund:
Don't know if this is still an issue (you'd have thought so, but nobody on second line complained about it recently?) – but even if so, the fix won't quite be this one. Closing.

PR #235, opened 2019-03-13
Allow Hub to log a hash of eidas user attributes

idrop:
Hub `saml-engine` will log a hash of eIDAS user attributes, in the case of an eIDAS Proxy Node originating journey through Hub. Logging is for security purposes.

An eIDAS-originating journey is evaluated by `saml-engine`  checking `config`, to see if the matching service entity id on the SAML response from the IDP is configured as an eIDAS proxy node.
You can see this at check at `IdpAuthnResponseTranslatorService:68`.

For an eIDAS journey, the matching service entity id will be the *request issuer id*, as this will be a non-matching journey, see`uk/gov/ida/hub/policy/services/AuthnResponseFromIdpService.java:54`.
This check occurs in `policy`, before the response is sent to `saml-engine`.

Proxy Node transaction yaml entries in Hub config will contain the data: `eidasProxyNode: true`.

The request-scoped `DecoratedSamlResponseToIdaResponseIssuedByIdpTransformer` will contain an Optional `EidasAttributesLogger`, which will log if it's present.

The merged PR that handles logging and hashing is https://github.com/alphagov/verify-saml-libs/pull/50


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/services/IdpAuthnResponseTranslatorService.java
```diff
@@ -38,24 +40,34 @@
...
+        this.eidasAttributesLogger = eidasAttributesLogger;
+        this.transactionsConfigProxy = transactionsConfigProxy;
     }
 
     public InboundResponseFromIdpDto translate(SamlAuthnResponseTranslatorDto samlResponseDto) {
         Response response = stringToOpenSamlResponseTransformer.apply(samlResponseDto.getSamlResponse());
         MdcHelper.addContextToMdc(response);
         try {
+            String matchingServiceEntityId = samlResponseDto.getMatchingServiceEntityId();
+            if (transactionsConfigProxy.isProxyNodeEntityId(matchingServiceEntityId)) {
```
idrop: Check with `config` whether this `matchingServiceEntityId` is an eidas proxy node service

vixus0: Assuming here the `matchingServiceEntityId` is the same as the transaction entity ID in a non-matching scenario? I don't really get why that `SamlAuthnResponseTranslatorDto` needs to know about the matching service but oh well.

idrop: Yes, updated PR message to note:

For an eIDAS journey, the matching service entity id will be the request issuer id, as this will be a non-matching journey, see uk/gov/ida/hub/policy/services/AuthnResponseFromIdpService.java:54.
This check occurs in policy, before the response is sent to saml-engine.

hub-saml/src/main/java/uk/gov/ida/saml/hub/domain/EidasAttributesLogger.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public void logEidasAttributesAsHash(Assertion assertion, Response response) {
+
+        Subject subject = assertion.getSubject();
+        String pid = subject.getNameID().getValue();
+        EidasResponseAttributesHashLogger hashLogger = loggerSupplier.get();
+        hashLogger.setPid(pid);
+
+        List<AttributeStatement> attributeStatements = assertion.getAttributeStatements();
+        List<Attribute> attributes = attributeStatements.get(0).getAttributes();
```
idrop: Worth noting that `attributeStatements` will list of exactly one `AttributeStatement`, as validated at
`uk/gov/ida/saml/core/validators/assertion/MatchingDatasetAssertionValidator.java:69`

PR #237, opened 2019-03-15
config resource "/transactions/{entityId}/is-eidas-proxy-node" to che…

idrop:
Add `config` resource `/transactions/{entityId}/is-eidas-proxy-node` which checks whether given `entityId` is configured as an eIDAS proxy node.

* returns 200 `true` for an entity that is configured as: `eidasProxyNode: true`
* return 200 `false` in all other cases


hub/config/src/main/java/uk/gov/ida/hub/config/domain/TransactionConfigEntityData.java
```diff
@@ -110,6 +110,10 @@
     @JsonProperty
     protected boolean usingMatching = true;
 
+    @Valid
+    @JsonProperty
+    protected boolean eidasProxyNode;
```
CharlesIC: I think we should have a default value of `false` for this. Otherwise all current transactions will have invalid config the moment this is merged since this is a non-nullable field. Just like `usingMatching ` above has a default of `true` to avoid this situation.

idrop: small b booleans are false by default, which is the required default state in this case. There is a test that checks this condition i.e. eidasProxyNode is never set to a value.

CharlesIC: Oh you're right! It will always be `false` by default. OK that looks then :)

PR #242, opened 2019-03-19
Replace debian packaging config with docker config

tlwr:
We do not need the debian packaging things anymore, so I have removed them.

Andy has copied all of the hub app configuration files out of the pipeline in alphagov/verify-infrastructure-config and put them in this repo.

A follow up PR will be raised removing the config files from the pipeline.


configuration/policy.yml
```diff
@@ -0,0 +1,66 @@
...
+      - type: access-logstash-console
+timeoutPeriod: 90m
+assertionLifetime: 60m
+acceptSelfSignedCerts: true
+matchingServiceResponseWaitPeriod: 60s
+sessionStore:
+  redis:
+    recordTTL: PT150m
+    uri: ${REDIS_HOST}
+samlEngineUri: https://saml-engine.${DOMAIN}:443
```
rhowe-gds: We could omit the explicit port 443 on HTTPS URLs, presumably

rhowe-gds: (this is not worth holding this PR up, just jotting down comments as I'm looking through the diff)

adityapahuja: I agree with @rhowe-gds. 443 is the default port number for HTTPS. It is safe to remove the explicit port number.

rhowe-gds: Actually this is already the case for saml-soap-proxy, for example
```
samlEngineUri: https://saml-engine.${DOMAIN}
configUri: https://config.${DOMAIN}
policyUri: https://policy.${DOMAIN}
```

PR #244, opened 2019-03-20
Add trust store metrics for all the trust stores

philandstuff:
PR #240 introduced metrics for hub and idp trust stores but missed out
these other ones.

saml-soap-proxy, saml-proxy, saml-engine and config all have "rp"
trust stores.  In addition, config has a "client" trust store - I'm
not sure if this is the best name for it but I don't know what else to
call it.


philandstuff:
I don't think i have access to the verify tech debt board - can somebody else add it please? @andy-paine

PR #249, opened 2019-03-28
AUTO: Add a splunk appender to SAML engine

andy-paine:
- We want to get our SAML Engine logs to Splunk but at the moment SAML
  engine logs to Logit via an egress proxy
- As there is no intermediate service such as Cloudwatch to bounce logs
  from into Splunk, we need to either get SAML Engine to send this stuff
to Splunk itself (this commit) or bounce logs from Logit
- This uses an external maven source that we would probably want to make
  a mirror from artifactory or build it ourselves

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,6 +56,7 @@ subprojects {
         }
         else {
           maven { url 'https://artifactory.ida.digital.cabinet-office.gov.uk/artifactory/whitelisted-repos' }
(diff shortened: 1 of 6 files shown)


andy-paine:
It would need to go inside the `if` section if we go down this road, yes. We would probably mirror the splunk repo through artifactory for "secure" builds. This was aimed at being PoC - the `SplunkAppenderFactory` could probably do with some unit tests and maybe a bit of error handling too.

PR #250, opened 2019-03-29
BAU: Fix keepAlive issue

adityapahuja:
The keepAlive value determines how long a persistent connection is kept open on the client side without being used (between two http requests). This value must be set less than the server side idleTimeout (defaults to 30 seconds in Dropwizard server) otherwise the connection can be closed on the server’s end resulting a org.apache.http.NoHttpResponseException when the client tries to reuse the connection.

SAML SOAP Proxy MSA client and SAML SOAP Proxy MSA HealthCheck client both have keepAlive set to 60 seconds. MSA server's idleTimeout defaults to 30 seconds. The clients' keepAlive are greater than the server's idleTimeout. There are 5,422 org.apache.http.NoHttpResponseException in SAML SOAP Proxy log file in production over the last 7 days via Kibana. This commit sets the keepAlive to 10 seconds instead of 60 seconds so that it is less than the server's idleTimeout. This change will remove org.apache.http.NoHttpResponseException.

This commit also sets SAML SOAP Proxy's idleTimeout to 70 seconds since MSA client has keepAlive to 60 seconds. This will also help remove org.apache.http.NoHttpResponseException on MSA side (not SAML SOAP Proxy side).

Author: @adityapahuja

### DIFF
--- a/configuration/saml-soap-proxy.yml
+++ b/configuration/saml-soap-proxy.yml
@@ -2,6 +2,7 @@ server:
   applicationConnectors:
     - type: http
       port: 8080
+      idleTimeout: 70 seconds
   requestLog:
     type: classic
     appenders:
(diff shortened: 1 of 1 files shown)


tlwr:
It is worth noting that we use the default value of keepalive in nginx which is 75 seconds which we do not override.

PR #252, opened 2019-03-29
GVzYvzsG: add response details on invalid session state

bjgill:
Where the session state is one that includes information on the IDP, decrypt the request and record the request details in the error created. Add tests for new function, and check that the old behaviour still works. Raises new exception whose message will get passed through to the frontend.

See https://trello.com/c/GVzYvzsG/104-understand-and-fix-invalidstate-errors

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/SessionRepository.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/SessionRepository.java
@@ -67,6 +67,12 @@ public class SessionRepository {
     }
 
     @Timed(name = Urls.SESSION_REPO_TIMED_GROUP)
+    public StateController getUnknownStateController(final SessionId sessionId) {
+        State currentState = getCurrentState(sessionId);
+        return controllerFactory.build(currentState, state -> dataStore.replace(sessionId, state));
+    }
+
+    @Timed(name = Urls.SESSION_REPO_TIMED_GROUP)
     public boolean sessionExists(SessionId sessionId) {
         return dataStore.hasSession(sessionId);
     }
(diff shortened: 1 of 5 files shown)


galund:
This (i.e. things like the IDP when this error occurs) feels like something we should be ensuring comes out into a separate field in our structured logs. This would enable us to filter and graph in Logit/Kibana properly? I've no idea how to do that mind you...

galund:
I also have a few reservations about committing this as-is, seeing as it only does the logging – which was acceptance criteria written on the assumption that we had already removed the check. (In other words, improving the logging is great, but most useful if we've also fixed the user experience.)

bjgill:
I agree that this isn't great. It looks like a good start, however. Could we get this reviewed and released, and then handle improving the UX in a separate PR to be released later today?

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/SessionRepository.java
```diff
@@ -66,6 +66,12 @@ public SessionId createSession(SessionStartedState startedState) {
         throw new InvalidSessionStateException(sessionId, expectedStateClass, currentState.getClass());
     }
 
+    @Timed(name = Urls.SESSION_REPO_TIMED_GROUP)
+    public StateController getUnknownStateController(final SessionId sessionId) {
```
galund: I think perhaps either `getFallbackStateController` or alternatively `getStateControllerRegardlessOfCurrentState`...

But it might be worth considering dropping this method, and just having `getStateController` allow `null` to be passed in as `expectedStateClass`. Your call.

bjgill: The latter proved quite hard, so I went for the former.

PR #256, opened 2019-04-02
Revert of GVzYvzsG: invalid session state logging

idrop:
Revert 4 commits related to GVzYvzsG as it causing 500 errors due to an IllegalStateException being thrown when state is currently 'Paused Registration', and user transitions to the same state.

The change in GVzYvzsG looks for a controller for the 'Paused Registration' state, which does not exist, and the exception is thrown.


jakubmiarka:
https://sentry.tools.signin.service.gov.uk/verify/prod-hub/issues/1281/

PR #257, opened 2019-04-04
BAU: use Java 11 in IntelliJ

bjgill:
Pre-requisite for starting to use Java 9+ features

PR #258, opened 2019-04-04
bau - dont build using travis with java8

JHjava:
- We run Java11 in prod, so there's no reason we need to be backwards compatible
- This will save time with the travis builds


bjgill:
Thanks - we'll probably want to land this around the same time as https://github.com/alphagov/verify-hub/pull/257.

PR #262, opened 2019-04-12
BAU Report coverage to Codacy

Wynndow:
https://trello.com/c/qPUHgpLn

This adds Gradle tasks and Travis config to calculate test coverage and
report it to Codacy.

The Jacoco plugin generates coverage data for each project after tests
are run. This data is merged into one file with a gradle task, and then
that data is converted to an XML and an HTML report with another task.

The XML is sent to Codacy, the HTML is available for humans if they want
it. The gradle tasks can be run locally without sending data to Codacy.

The encrypted var in the travis file is CODACY_PROJECT_TOKEN and is
taken from the codacy-bot account that we have.

Screen short of an HTML report:
_Users_christopherwynne_GDS_verify-hub_build_reports_jacoco_jacocoRootReport_html_index html(iPad Pro)


CharlesIC:
Tested locally. Work on my machine 👍

PR #266, opened 2019-04-25
use latest verify-saml-libs in Hub, built with OpenJDK 11

idrop:
We want to standardise Hub compile and runtimes to OpenJDK 11.

This PR will allow Hub to depend on the latest `verify-saml-libs`, compiled with with OpenJDK 11.


idrop:
Please ignore commit https://github.com/alphagov/verify-hub/pull/266/commits/48b18b14fe0050ef45824f7521d834e31361148c, that was meant to be run on Proxy Node

PR #277, opened 2019-05-08
l4zBlwLF: Add config options for self SelfServiceConfig

kerrr:
* Added self service section to the configuration of the config service
  itself to configure enabling self service globaly and provide the source
  URL that the self service data will be availible from.

* Added selfService option to the transaction config, so enable self
  service for that service.


kerrr:
I'm thinking that these config setting are specific to self service, i.e. enable self service specifically. The code for loading config / certificates from a remote source would of course be generic, not mentioning self service.

PR #282, opened 2019-05-14
ZD3679369: record unhealthy MSA info

bjgill:
Ensure that MSAs that have been unhealthy for a long time still appear on the Grafana dashboard. Some info fields will be empty (e.g. version), but that's probably better than nothing.

See https://govuk.zendesk.com/agent/tickets/3679369


bjgill:
It should be fine, but I don't think there's a way to find out for sure without pushing to prod and seeing what becomes of the yak dashboard. I'm going to merge now - please shout if the yak dashboard does anything silly.

PR #289, opened 2019-05-23
l4zBlwLF Implemented caching on the remote config.

j-maynard:
Implemented caching on the remote config and setup the getConfig
method to throw IOEXceptions which can be caught to fallback to
the normal fed-config.


hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -17,15 +18,23 @@
 import java.io.InputStream;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.Date;
 
 public class S3ConfigSource {
     private static final Logger LOG = LoggerFactory.getLogger(S3ConfigSource.class);
     private ConfigConfiguration configConfiguration;
-    private AmazonS3 s3Client;
+    private S3Object s3Object;
```
kerrr: This doesn't look right to me. Does this work? Shouldn't we be getting the s3Object every request.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -17,15 +18,23 @@
...
+    private long cacheLengthInMillis;
+    private Date lastModifiedDate;
+    private Long lastAccessedTime;
+    private RemoteConfigCollection currentConfig;
 
     public S3ConfigSource(ConfigConfiguration configConfiguration, AmazonS3 s3Client) {
         this.configConfiguration = configConfiguration;
-        this.s3Client = s3Client;
+        this.cacheLengthInMillis = configConfiguration.getSelfService().getCacheLengthInSeconds() * 1000;
+        this.s3Object = s3Client.getObject(
```
kerrr: see above comment. This feels wrong.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -49,33 +58,61 @@ public static S3ConfigSource setupS3ConfigSource(ConfigConfiguration configConfi
...
         } finally {
-            if(fullObject != null) {
+            if(this.s3Object != null) {
                 try {
-                    fullObject.close();
+                    this.s3Object.close();
                 } catch(IOException e) {
                     throw new RuntimeException(e);
                 }
             }
```
kerrr: I think we could do a better job of the exception handling here. If we're going to throw the IOException anyway, they we shouldn't need all this.

chrisholmes:
We've got existing prior art for using a [`LoadingCache`](https://github.com/google/guava/wiki/CachesExplained#from-a-cacheloader) for this kind of stuff. It should make this a lot simpler and we won't have to think about concurrency isssue so much.

jakubmiarka:
Should this be closed in favour of https://github.com/alphagov/verify-hub/pull/300 ?

kerrr:
Yes. It's been superseded.

PR #290, opened 2019-05-24
N7edPMiO: Improve MSA Logging

chrisclayson:
## What

MSA logs don't contain a lot of useful information. It's hard to use them to identify what a problem is.

We should implement error logging changes to the MSA so that the full stacktrace is returned to the Verify Hub when an HTTP 500 error happens in the MSA.

There is some further background in Anshul's [investigation into MSA 500 errors](https://docs.google.com/document/d/10-oAqzKiPXu4nzwD9GL7knABwKxOLNY4oUlo6gC4uRA/edit#).

Original Zendesk ticket:
https://govuk.zendesk.com/agent/tickets/3651026

## How
Change special exception handling to look for exception name at start of message, as message will potentially now contain full stack trace rather than just contain the exception name.

## Related

https://github.com/alphagov/verify-matching-service-adapter/pull/122

### DIFF
--- a/hub/saml-soap-proxy/src/main/java/uk/gov/ida/hub/samlsoapproxy/client/AttributeQueryRequestClient.java
+++ b/hub/saml-soap-proxy/src/main/java/uk/gov/ida/hub/samlsoapproxy/client/AttributeQueryRequestClient.java
@@ -78,7 +78,7 @@ public class AttributeQueryRequestClient {
 
                 // The MSA sometimes returns 500s with a singularly unhelpful body. Special case handling to get Sentry
                 // to split out these errors. See https://trello.com/c/N7edPMiO/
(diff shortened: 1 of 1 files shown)


chrisclayson:
[Improve MSA logging](https://trello.com/c/N7edPMiO/52-improve-msa-logging)

robinmitra:
Not entirely clear from where this exception is being thrown:

`uk.gov.ida.matchingserviceadapter.exceptions.LocalMatchingServiceException`

chrisclayson:
> uk.gov.ida.matchingserviceadapter.exceptions.LocalMatchingServiceException

It's not anymore, that was part of our first solution to this change, which we abandoned. If you look at the final version of this PR, references to that exception have now been removed!

chrisclayson:
Updated PR description to reflect change of approach.

PR #293, opened 2019-05-31
l4zBlwLF: Use remote config to overwrite local one

jakubmiarka:


kerrr:
The scope of this PR has got a little bit lost. The original intent was for this to create a uniform ConfigRepository layer for Matching Services config and Transaction config. These are called by the relevant Resources and handle merging local config with any remote config retrieved from the S3ConfigSource.

We need to do some more work on the S3ConfigSource and integration tests for it, but this is outside the scope for this PR.

jakubmiarka:
[Hub config service can override local config with self service metadata](https://trello.com/c/EI9xiWXp/393-hub-config-service-can-override-local-config-with-self-service-metadata)

PR #295, opened 2019-06-07
BAU: update AssertJ

kerrr:


JHjava:
Failing to compile. This needs to be removed `import static org.assertj.guava.api.Assertions.assertThat;
` from AttributeQueryGeneratorTest.java.

PR #297, opened 2019-06-07
l4zBlwLF: Simplify generics for CertificateService

kerrr:
Everything that implements `CertificateConfigurable` also implements
`EntityIdentifiable`. In `CertificateService` we had a generic type
definition that required both implementations. We can simplify this by
having `CertificateConfigurable` extend `EntityIdentifiable`.


kerrr:
This is all local to the Config Service. :)

PR #299, opened 2019-06-10
port AttributeQuery request Timer to prometheus

philandstuff:
We have a Dropwizard Metrics Timer to record times taken for
AttributeQuery requests (ie SOAP requests to matching services).
However they are somewhat fiddly to deal with in prometheus.

This creates a prometheus Summary to capture this data.  A big
advantage is that the matching service URI will be available as a
label rather than a weirdly formatted part of the metric name, so we
can more easily do queries for things like "show me the slowest 3
matching service URIs".

I picked the quantiles to be a useful subset of the current quantiles
on the dropwizard Timer.


philandstuff:
:thinking: that might actually be a problem. I’ll check in the morning. I don’t currently have a working java 11 so I couldn’t run tests locally.

philandstuff:
We discussed face to face and realised it's not a problem because the field is `static`

PR #300, opened 2019-06-11
l4zBlwLF Implemented Guava Cache

j-maynard:
This code implements a basic S3Object cache using Guava caching.


hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -37,12 +38,14 @@ public S3ConfigSource(ConfigConfiguration configConfiguration, AmazonS3 s3Client
         CacheLoader<String, RemoteConfigCollection> loader;
         loader = new CacheLoader<String, RemoteConfigCollection>() {
             private RemoteConfigCollection currentConfig;
-            private Date lastModified;
 
             @Override
             public RemoteConfigCollection load(String key) {
                 S3Object s3Object = getS3Object(key);
-                if (this.lastModified.before(s3Object.getObjectMetadata().getLastModified())) {
+                if(lastModified == null) {
```
chrisholmes: If you use `CacheLoader` then you can use [`expireAfterWrite(...)`] and not need to use `lastModified`. Every time the cache expires for that key the next read will force the cache to load a new value.

Your code could look something like this:
```
  LoadingCache<Key, RemoteConfigCollection> graphs = CacheBuilder.newBuilder()
       .maximumSize(1) //optional
       .expireAfterWrite(TIMEOUT, TimeUnit.MINUTES)
       .build(
           new CacheLoader<Key, RemoteConfigCollection>() {
             public Graph load(Key key) throws AnyException {
               return s3client.getConfigFromS3();
             }
           });
```

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -37,12 +38,14 @@ public S3ConfigSource(ConfigConfiguration configConfiguration, AmazonS3 s3Client
...
+                    updateObjectData(getS3Object(key));
```
chrisholmes: I think now might be the time to extract the s3 client code into its own class as there's a lot going on here now.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: I'm looking at the expireAfterWrite stuff now.

chrisholmes: Thanks, here's an example from the hub: https://github.com/alphagov/verify-hub/blob/7109a388a896eb8c3aa8fbf303a490b05c8b4b2c/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/config/CertificatesConfigProxy.java#L29

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -1,81 +1,122 @@
...
-        this.s3Client = s3Client;
+        this.selfServiceConfig = configConfiguration.getSelfService();
+        if (selfServiceConfig != null && selfServiceConfig.isEnabled()){
+            this.s3Client = s3Client;
+            this.objectMapper = new ObjectMapper();
+            objectMapper.setDateFormat(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss"));
+
+            CacheLoader<String, RemoteConfigCollection> cacheLoader = new S3ConfigCacheLoader();
+            long expiry = selfServiceConfig.getCacheExpiryInSeconds();
+            TimeUnit unit = TimeUnit.SECONDS;
```
chrisholmes: Dropwizard has support for doing these type conversion using the Duration type. See [here](https://github.com/alphagov/verify-hub/blob/master/hub/policy/src/main/java/uk/gov/ida/hub/policy/configuration/PolicyConfiguration.java#L31) for an example

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -1,81 +1,122 @@
...
+            } catch(IOException e) {
+                LOG.error("An error occurred trying to get or process object {} from S3 Bucket {}",
+                        selfServiceConfig.getS3ObjectKey(),
+                        selfServiceConfig.getS3BucketName(), e);
+                throw(e);
+            }
+        }
+
+        @Override
+        public ListenableFuture<RemoteConfigCollection> reload(String key, RemoteConfigCollection existingConfig) {
```
chrisholmes: I think you're making this asynchronous at the wrong point. You want the most expensive operation, the network request to S3, to be made asynchronous or you would not get any benefit out of overriding this method. The asynchronous task should start at the top of this method before the S3 call.

Also, given how much cheap it is to move small amount of data around in memory compared to a network request you might want to consider ignoring all the lastModified logic.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -1,81 +1,122 @@
...
+    private LoadingCache<String, RemoteConfigCollection> cache;
+    private ObjectMapper objectMapper;
 
     public S3ConfigSource(ConfigConfiguration configConfiguration, AmazonS3 s3Client) {
-        this.configConfiguration = configConfiguration;
-        this.s3Client = s3Client;
+        this.selfServiceConfig = configConfiguration.getSelfService();
+        if (selfServiceConfig != null && selfServiceConfig.isEnabled()){
+            this.s3Client = s3Client;
+            this.objectMapper = new ObjectMapper();
```
chrisholmes: Why not use the one injected by guice?

Alternatively, when we create ObjectMappers we tend to use `Jackson.newObjectMapper`.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -1,81 +1,122 @@
...
+    public static final RemoteConfigCollection EMPTY_COLLECTION = new RemoteConfigCollection(null, null, EMPTY_CONNECTED_SERVICE_CONFIG_MAP, EMPTY_MATCHING_SERVICE_CONFIG_MAP, EMPTY_SERVICE_PROVIDER_CONFIG_LIST);
+    private SelfServiceConfig selfServiceConfig;
     private AmazonS3 s3Client;
+    private LoadingCache<String, RemoteConfigCollection> cache;
+    private ObjectMapper objectMapper;
 
     public S3ConfigSource(ConfigConfiguration configConfiguration, AmazonS3 s3Client) {
-        this.configConfiguration = configConfiguration;
-        this.s3Client = s3Client;
+        this.selfServiceConfig = configConfiguration.getSelfService();
```
chrisholmes: Why not just inject `SelfServiceConfig`?

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -1,81 +1,122 @@
...
+        if (selfServiceConfig != null && selfServiceConfig.isEnabled()){
+            this.s3Client = s3Client;
+            this.objectMapper = new ObjectMapper();
+            objectMapper.setDateFormat(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss"));
```
chrisholmes: How is this different from the standard date format?

hub/config/src/main/java/uk/gov/ida/hub/config/configuration/SelfServiceConfig.java
```diff
@@ -31,6 +31,9 @@
     @JsonProperty
     private String s3SecretKeyId;
 
+    @JsonProperty
+    private long cacheExpiryInSeconds;
```
chrisholmes: See comment below. This can be a `Duration`.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -1,81 +1,122 @@
...
-            BasicAWSCredentials awsCreds = new BasicAWSCredentials(configConfiguration.getSelfService().getS3AccessKeyId(),
-                    configConfiguration.getSelfService().getS3SecretKeyId());
-            s3Client = AmazonS3ClientBuilder.standard()
-                    .withRegion(configConfiguration.getSelfService().getAwsRegion())
-                    .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
-                    .build();
-        } else {
-            s3Client = AmazonS3ClientBuilder.defaultClient();
+    public RemoteConfigCollection getRemoteConfig(){
+        if ( selfServiceConfig == null || !selfServiceConfig.isEnabled()){
```
chrisholmes: Maybe selfServiceConfig should be optional?

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
```diff
@@ -1,81 +1,122 @@
...
-                    fullObject.close();
-                } catch(IOException e) {
-                    throw new RuntimeException(e);
-                }
-            }
+            inputStream.close();
         }
     }
 
+    private S3Object getS3Object(String key) {
```
chrisholmes: There's alot of S3 client and object code scattered about. You could create a class that wraps the S3Client and the Objectmapper that behaves similar to our http clients. I think it would make our code a lot simpler if we could do something like:

```
client.get(SelfServiceMetadata.class)
```

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: Oh, nice. I've used that.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: OK, I've made the whole lot async. I think the lastModified is a useful optimisation.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: It was just there before. Not sure why it's been specified explicitly. I've removed it and everything seems to still work.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: Thinking about it, we should maybe default it to be there, with enabled false.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: Can we have a think about that after this PR?

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
chrisholmes: Yeah, I think that's possible with DW's configuration

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
chrisholmes: In that case, you should be able to inject the app-scope ObjectMapper instance.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: OK, done that.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
kerrr: Doing that now.

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
chrisholmes: sure

hub/config/src/main/java/uk/gov/ida/hub/config/data/S3ConfigSource.java
chrisholmes: You need to include L100 in the Future as that's the most expensive action.

kerrr: OK, I'd miss-understood what was happening here.

kerrr: Done

PR #305, opened 2019-06-17
OGD-169: Move hub-specific code from verify-utils-libs into verify-hub

mrwilson:
The `eventsink` package in verify-utils-libs has been moved here, as the hub is the only place that uses it.

It also breaks the dependency in verify-utils-libs on verify-event-emitter, as the hub already pulls this dependency in.

First commit creates a new subproject (`:hub:shared`) for shared components.
Second commit replaces imports from verify-utils-libs with the local library.


JHjava:
There's now 2 EventSinkHubEvent.java in verify-hub. There's one in Policy and one in shared. The one in Policy should probably be removed and whatever is using it should use the shared class?

mrwilson:
@JHjava the EventSinkHubEvent in policy seems to be a deliberate duplicate, as the `SessionId` it references is also from the policy module.

In order to remove it from `policy` and use the one in `shared`, you'd need to move `SessionId` too else you end up with a circular dependency shared -> policy -> shared ...

We can:

(a) Rename the `EventSinkHubEvent` in `policy` to make it clear it's specific to that implementation of SessionId
(b) Move `SessionId` into `shared` too, to break the circular dependency.
(c) Use the `SessionId` in verify-utils-libs everywhere in verify-hub and remove the instance in `policy`, although they are very slightly different (the one in verify-utils-libs has `@JsonValue(true)` on its `.toString()`

IMO, (c) feels the better option, it's getting rid of duplicated code between libs and hub.

JHjava:
Good spot, Alex. Let's do that in a separate PR.

mrwilson:
We're reading the `sessionId` query param into `SessionId` objects in the code - as long as the implementations are the same, I think it should be okay? Happy to take a second opinion on this.

I agree, separate PR.

PR #307, opened 2019-06-18
OGD-169: Use :hub:shared inside Dockerfile (fix)

mrwilson:
* The build passed locally and in Travis, but failed in Concourse because the new library from previous commit was not included in the Dockerfile build.

Co-authored-by: James Howes <james.howes@digital.cabinet-office.gov.uk>


mrwilson:
Thanks, all.

PR #308, opened 2019-06-18
EID-1509 Update the list of supported MSA versions

Wynndow:
At the time of writing there are only two versions being used by RPs. We
don't expect RP's to downgrade so we can reduce the number of supported
versions.

The latest release (4.0.0) is added here.

null and 0 are left as it's assumed they are valid.

This list is used by the Jenkins job [ida-msa-backward-compatibility-check](https://build.ida.digital.cabinet-office.gov.uk/job/ida-msa-backward-compatibility-check/).


CharlesIC:
Some integration tests need updating

Wynndow:
Yep, that's what you get for not running pre-commit. Have fixed.

PR #314, opened 2019-06-28
Parse Self Service enabled config

brenetic:
Set the other required configuration for the config service to read the new metadata bucket.

Corresponding PRs:
https://github.com/alphagov/verify-infrastructure-config/pull/158
https://github.com/alphagov/verify-infrastructure/pull/244

https://trello.com/c/glz5vLuk/485-update-the-config-in-hub-to-enable-the-e2e-journey

### DIFF
--- a/configuration/config.yml
+++ b/configuration/config.yml
@@ -31,3 +31,8 @@ certificateExpiryDateCheckServiceConfiguration:
   enable: true
 certificateOcspRevocationStatusCheckServiceConfiguration:
   enable: true
+selfServiceConfiguration:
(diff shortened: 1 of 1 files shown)


configuration/config.yml
```diff
@@ -31,3 +31,8 @@ certificateExpiryDateCheckServiceConfiguration:
   enable: true
 certificateOcspRevocationStatusCheckServiceConfiguration:
   enable: true
+selfServiceConfiguration:
```
kerrr: This should be `selfService`
```suggestion
selfService:
```

PR #322, opened 2019-07-12
xHGobHw4: Use enabled flag in TrustStore Config

kerrr:
We want to be able to turn off validating certificate chains with our
root trust store for RPs (and their MSAs). We have previously added an
`enabled` flag to the `ClientTrustStoreConfiguration` in
`verify-utils-libs`.

This commit:

Updates `TrustStoreForCertificateProvider` to return `Optional<KeyStore>`
from `getTrustStoreFor`.

Uses the new `enabled` flag in `ClientTrustStoreConfiguration` to
determine whether or not to return the configured trust store.

Updates `ConfigServiceKeyStore` to validate using the trust store only
if it is present. i.e. the Optional returned from `getTrustStoreFor` is
not empty.

co-authored-by: Alan Carter <alan.carter@digital.cabinet-office.gov>

### DIFF
--- a/hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/config/ConfigServiceKeyStoreTest.java
+++ b/hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/config/ConfigServiceKeyStoreTest.java
@@ -1,5 +1,6 @@
 package uk.gov.ida.hub.samlengine.config;
 
+import com.google.common.collect.ImmutableList;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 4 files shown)


kerrr:
I'm going to ignore the Codacy fail for naming, going with consistency on this one.

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/config/ConfigServiceKeyStoreTest.java
```diff
@@ -121,10 +123,20 @@ public void getVerificationKeyForEntity_shouldThrowExceptionIfCertificateIsInval
     }
 
     @Test
-    public void getEncryptionKeyForEntity_shouldGetEncryptionKeysFromConfigCertificateProxy() throws Exception {
+    public void getVerificationKeyForEntity_shouldNotValidateWhenTrustStoreDisabled() {
```
jakubmiarka: 👍

kerrr:
Although the tests in here refer to IDP Certs, this is not the pathway that IDP certs get verified with. (I think they are just using any old valid cert.) The `TrustStoreForCertificateProvider` that we mock out throws an exception if you call `getTrustStoreFor(FederationEntityType.IDP)`. I'm not sure how it ended up like this. It's either something old that is left over, or was there in expectation of future code that never materialised.

Given that we should strip this all out after we have got it all disabled in prod, I'm a bit loathed to tests paths that can never be executed.

PR #325, opened 2019-07-16
0BFh8hBN: RP truststore feature flag in saml-proxy

alan-gds:
This is the saml-proxy equivalent of https://github.com/alphagov/verify-hub/pull/322


hub/saml-proxy/src/test/java/uk/gov/ida/hub/samlproxy/config/ConfigServiceKeyStoreTest.java
```diff
@@ -54,7 +57,7 @@
     private ConfigServiceKeyStore configServiceKeyStore;
 
     @Before
-    public void setup() throws CertificateException {
+    public void setup() {
```
jakubmiarka: According to Codacy
```suggestion
    public void setUp() {
```

PR #342, opened 2019-08-05
Simplify our handing of arrays in various places

rhowe-gds:
Make use of `singletonList()` where we can
Stop passing sizes to `toArray()` where we don't need to
Simplify `CertificateService` by removing the unused `getAllCertificates` method and make use of the constructor to `ArrayList` which takes a `List` in `getAllCertificates`

### DIFF
--- a/hub/config/src/integration-test/java/uk/gov/ida/integrationtest/hub/config/apprule/support/ConfigAppRule.java
+++ b/hub/config/src/integration-test/java/uk/gov/ida/integrationtest/hub/config/apprule/support/ConfigAppRule.java
@@ -30,7 +30,7 @@ import java.util.function.Supplier;
 import java.util.stream.IntStream;
 
 import static io.dropwizard.testing.ConfigOverride.config;
-import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
 import static uk.gov.ida.hub.config.domain.builders.CountryConfigBuilder.aCountryConfig;
 import static uk.gov.ida.hub.config.domain.builders.IdentityProviderConfigDataBuilder.anIdentityProviderConfigData;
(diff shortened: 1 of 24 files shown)


hub/config/src/integration-test/java/uk/gov/ida/integrationtest/hub/config/apprule/support/ConfigAppRule.java
```diff
@@ -77,7 +77,7 @@ public ConfigAppRule(ConfigOverride... configOverrides) {
                 .add(config("translationsDirectory", TRANSLATIONS_RELATIVE_PATH))
                 .add(configOverrides)
                 .build();
-        return overrides.toArray(new ConfigOverride[overrides.size()]);
+        return overrides.toArray(new ConfigOverride[0]);
```
kerrr: Why not initialise the array to be the correct size. Otherwise it ends up creating a new one of the correct size anyway.

rhowe-gds: https://shipilev.net/blog/2016/arrays-wisdom-ancients/

kerrr: Isn't that interesting.

hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateService.java
```diff
@@ -83,6 +75,6 @@ private T getConfig(String entityId){
                 .findFirst()
                 .orElseThrow(NoCertificateFoundException::new)
                 .get(entityId)
-                .get();
+                .orElseThrow();
```
jakubmiarka: 👍

PR #343, opened 2019-08-05
Simplify UnknownMethodAlgorithmLogger

rhowe-gds:
We don't need the static initialiser here now that we can do `Set.of(...)`

Also, by passing `Signature` objects around a bit more, we can reduce some boilerplate and remove an `Optional` at the same time as making better use of the `Optional`s we have

### DIFF
--- a/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/logging/UnknownMethodAlgorithmLogger.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/logging/UnknownMethodAlgorithmLogger.java
@@ -7,6 +7,7 @@ import org.apache.xml.security.signature.XMLSignature;
 import org.opensaml.saml.saml2.core.Assertion;
 import org.opensaml.saml.saml2.core.AuthnRequest;
 import org.opensaml.saml.saml2.core.Response;
+import org.opensaml.xmlsec.algorithm.DigestAlgorithm;
 import org.opensaml.xmlsec.algorithm.descriptors.DigestSHA256;
 import org.opensaml.xmlsec.algorithm.descriptors.DigestSHA512;
 import org.opensaml.xmlsec.signature.Signature;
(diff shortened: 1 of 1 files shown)


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/logging/UnknownMethodAlgorithmLogger.java
```diff
@@ -37,98 +37,73 @@
...
 
     public static void probeResponseForMethodAlgorithm(final InboundResponseFromIdp response) {
         if (response != null) {
-            final Optional<Signature> signature = response.getSignature();
-            if (signature != null) {
-                final String signatureMethodAlgorithm = getSignatureMethodAlgorithm(signature);
-                final String digestMethodAlgorithm = getDigestMethodAlgorithm(signature);
-                logMethodAlgorithm(Role.IDP, signatureMethodAlgorithm, digestMethodAlgorithm, Response.DEFAULT_ELEMENT_LOCAL_NAME);
-            }
+            response.getSignature().ifPresent((signature) ->
```
rhowe-gds: Maybe this could throw an NPE now actually

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/logging/UnknownMethodAlgorithmLogger.java
```diff
@@ -37,98 +37,73 @@
...
 
     public static void probeAuthnRequestForMethodAlgorithm(final AuthnRequestFromRelyingParty authnRequest) {
         if (authnRequest != null) {
-            final Optional<Signature> signature = authnRequest.getSignature();
-            if (signature != null) {
-                final String signatureMethodAlgorithm = getSignatureMethodAlgorithm(signature);
-                final String digestMethodAlgorithm = getDigestMethodAlgorithm(signature);
-                logMethodAlgorithm(Role.SP, signatureMethodAlgorithm, digestMethodAlgorithm, AuthnRequest.DEFAULT_ELEMENT_LOCAL_NAME);
-            }
+            authnRequest.getSignature().ifPresent((signature) ->
```
rhowe-gds: Maybe this could throw an NPE now :(

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/logging/UnknownMethodAlgorithmLogger.java
idrop: Think the `response` signature is validated prior to this call so `response.getSignature()` should be null safe

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/logging/UnknownMethodAlgorithmLogger.java
idrop: as above, a `SamlTransformationErrorException` would be thrown if signature was invalid

PR #351, opened 2019-08-09
Replace guava Predicates with Java ones

rhowe-gds:
Does this actually work?


kerrr:
Can't a lot of this stuff just be removed? Why do we need the class `EnabledIdpPredicate` with it's associated `EnabledIdpPredicateTest` at all? Might have a look and see if I can get rid of all this.

rhowe-gds:
> Can't a lot of this stuff just be removed? Why do we need the class `EnabledIdpPredicate` with it's associated `EnabledIdpPredicateTest` at all? Might have a look and see if I can get rid of all this.

Quite possibly. Knock yourself out!

kerrr:
We can definitely just get rid of a lot of this. Some is straightforward to replace, but other bits are a bit inscrutable, so I didn't have time to make sure it was all correct. It's worthwhile getting rid of the guava now though, so I'm approving and I'll take another run at it when I've got time.

PR #354, opened 2019-08-12
BMqfOfOf: Stop validating self-service certs against trust chain

alan-gds:
https://trello.com/c/BMqfOfOf/542-stop-validating-certificates-against-federation-certificate-chain-when-serving

### DIFF
--- a/hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateService.java
+++ b/hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateService.java
@@ -21,7 +21,7 @@ import static java.util.stream.Collectors.partitioningBy;
 
 public class CertificateService <T extends CertificateConfigurable<T>> {
 
-    private static final Logger LOG = LoggerFactory.getLogger(ManagedEntityConfigRepository.class);
+    private static final Logger LOG = LoggerFactory.getLogger(CertificateService.class);
 
     private final ManagedEntityConfigRepository<T> connectedServiceConfigRepository;
     private final ManagedEntityConfigRepository<T> matchingServiceConfigRepository;
@@ -63,8 +63,11 @@ public class CertificateService <T extends CertificateConfigurable<T>> {
                 .stream()
                 .collect(partitioningBy(cd -> certificateValidityChecker.isValid(cd)));
 
-        certsByValidity.get(false)
-                .forEach(cd -> LOG.warn("Signature verification certificates were requested for entityId '{}' but at least one is invalid", entityId));
+        int numberOfBadCerts = certsByValidity.get(false).size();
+        String isOrAre = numberOfBadCerts == 1 ? "is" : "are";
+        if (numberOfBadCerts > 0) {
+            LOG.warn("Signature verification certificates were requested for entityId '{}'; {} of them {} invalid", entityId, numberOfBadCerts, isOrAre);
+        }
 
         return certsByValidity.get(true)
                 .stream()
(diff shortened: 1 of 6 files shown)


hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateService.java
```diff
@@ -50,6 +55,12 @@ public  Certificate encryptionCertificateFor(String entityId) {
 
     public List<Certificate> signatureVerificationCertificatesFor(String entityId) {
         T config = getConfig(entityId);
+
+        config.getSignatureVerificationCertificates()
+                .stream()
+                .filter(cd -> !certificateValidityChecker.isValid(cd))
+                .forEach(cd -> LOG.warn("Signature verification certificates were requested for entityId '{}' but at least one is invalid", entityId));
+
```
kerrr: We could call `getSignatureVerificationCertificates().stream()` once then partition on `isValid()`. We could then handle the true case and the false case. Might be nicer?

kerrr:
Looking good. I think we need to test that we are logging in the CertificateService correctly and we'll need an integration test to exercise the this. I can look at the integration test if your having issues with them.

hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateService.java
```diff
@@ -50,9 +58,16 @@ public  Certificate encryptionCertificateFor(String entityId) {
...
     public List<Certificate> signatureVerificationCertificatesFor(String entityId) {
         T config = getConfig(entityId);
-        return config.getSignatureVerificationCertificates()
+
+        Map<Boolean, List<Certificate>> certsByValidity = config.getSignatureVerificationCertificates()
+                .stream()
+                .collect(partitioningBy(cd -> certificateValidityChecker.isValid(cd)));
+
+        certsByValidity.get(false)
+                .forEach(cd -> LOG.warn("Signature verification certificates were requested for entityId '{}' but at least one is invalid", entityId));
```
kerrr: I think we could word the message better now. We know how many of the certs are invalid at this stage. We could also make sure we don't log more than one message.
something like:
```
int badCertCount = certsByValidity.get(false).size();
        if (badCertCount > 0){
            LOG.warn("Signature verification certificates were requested for entityId '{}' {} invalid", entityId, badCertCount == 1 ? "one is" : "both are");
        }
```

hub/config/src/test/java/uk/gov/ida/hub/config/application/CertificateServiceTest.java
```diff
@@ -288,4 +324,14 @@ public void getAllCertificatesReturnsAllTransactionAndMatchingServiceCertificate
         assertThat(actualCertificateSet).containsAll(expectedCertificateSet);
     }
 
+    private void checkForExpectedLogWarnings(Iterable<String> expectedMessages) {
```
alan-gds: This is slight overkill as all of the tests in fact check for exactly one expected message (or don't call the check), but while writing it I was expecting there'd be some cases where we'd need to handle several.  I think the overengineered version will be more resilient against future changes than a simple .get(0) version would be, though, so having written it, I'd prefer to keep it.

hub/config/src/main/java/uk/gov/ida/hub/config/application/CertificateService.java
```diff
@@ -21,7 +21,7 @@
 
 public class CertificateService <T extends CertificateConfigurable<T>> {
 
-    private static final Logger LOG = LoggerFactory.getLogger(ManagedEntityConfigRepository.class);
+    private static final Logger LOG = LoggerFactory.getLogger(CertificateService.class);
```
kerrr: 😅👍🏻

hub/config/src/test/java/uk/gov/ida/hub/config/application/CertificateServiceTest.java
kerrr: Yup, LGTM.

PR #362, opened 2019-08-19
s4ekDNZR: Unit test for OcspCertificateChainValidationService

alan-gds:
* Checks that gauges are updated based on cert chain validity.

https://trello.com/c/s4ekDNZR/558-ocspcertificatechainvalidationservice-should-use-certificates-validation-flag-to-determine-validation-method

### DIFF
--- a/hub/config/src/test/java/uk/gov/ida/hub/config/application/OcspCertificateChainValidationServiceTest.java
+++ b/hub/config/src/test/java/uk/gov/ida/hub/config/application/OcspCertificateChainValidationServiceTest.java
@@ -1,0 +1,112 @@
+package uk.gov.ida.hub.config.application;
+
+import io.prometheus.client.Gauge;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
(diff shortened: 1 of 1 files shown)


brenetic:
Should they be three separate tests? Or do Java tests have things like `contexts` for grouping similar tests which may have slightly separate expectations?

alan-gds:
> Should they be three separate tests? Or do Java tests have things like `contexts` for grouping similar tests which may have slightly separate expectations?

I don't really see this as three tests in one, I see it as testing a single series of operations, with assertions before and after each step so that it's clear that things are proceeding as expected.  The gauges are stateful, so it's not enough to just test one single operation.

I could split it into three tests (check that all gauges start at zero, check that gauges get updated to 1 when appropriate, and check that gauges are set back from 1 to 0 if a cert is invalid when checked a second time).  For the last of those three tests, I'd still want to assert that the gauges were set to 1 before running the service a second time, which would basically be duplicating the second test...so it just felt better to do them all together.

I'm willing to change it if you twist my arm.

PR #366, opened 2019-08-30
EID-1608: Pass original SAML message if assertions unsigned

JohnWattsGDS:
To preserve message integrity messages with unsigned assertions must be passed in their original form so that the country's signature can be checked. So:

When translating, use validateEidas method in ResponseAssertionsFromCountryValidator#validate so that unsigned assertions are valid.

Check whether any assertions are unsigned in InboundResponseFromCountry#translate(SamlAuthnResponseTranslatorDto) and set flag in the InboundResponseFromCountry accordingly and also attach original saml message to response in either case.  Other components can deal with this response as needed.

Add tests (which could be refactored at some point in the future to remove duplication).


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/domain/InboundResponseFromCountry.java
```diff
@@ -10,6 +12,10 @@
     private Optional<String> statusMessage;
     private Optional<String> encryptedIdentityAssertionBlob;
     private Optional<LevelOfAssurance> levelOfAssurance;
+    @JsonProperty
```
idrop: I'm unsure why this boolean property alone has to be annotated as a `@JsonProperty`. Could it be removed if we renamed the property to`assertionsUnsigned`?

JohnWattsGDS: I think it's more down to the getter's name than the property's; I may have been able to annotate the getter instead of the property but it serves as a better warning sticking out at the top.

Because of the dark magic used in the bowels of Jackson (or whatever library it is), the getter will be ignored unless it's called `isAssertionsUnsigned` and given that possibility, I'd rather take the annotation - a) because it sounds ugly in my head when I read it and b) because there's a chance it could be misread as `isAssertionUnsigned` which is a thing in verify-saml-libs that's used in a stream operation to set the property ...

It's all very confusing but the take home is it took me a lot longer than I'd hope it would to smoke the "bug" out!

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/services/CountryAuthnResponseTranslatorService.java
```diff
@@ -101,11 +127,14 @@ private InboundResponseFromCountry toModel(ValidatedResponse response, Optional<
...
 
         return new InboundResponseFromCountry(
-            response.getIssuer().getValue(),
-            validatedIdentityAssertionOptional.map(Assertion::getSubject).map(Subject::getNameID).map(NameID::getValue),
-            Optional.ofNullable(status).map(CountryAuthenticationStatus::getStatusCode).map(CountryAuthenticationStatus.Status::name),
-            status.getMessage(),
-            passthroughAssertion.map(assertion -> assertionBlobEncrypter.encryptAssertionBlob(matchingServiceEntityId, assertion.getUnderlyingAssertionBlob())),
-            levelOfAssurance);
+                response.getIssuer().getValue(),
+                validatedIdentityAssertionOptional.map(Assertion::getSubject).map(Subject::getNameID).map(NameID::getValue),
```
tomrosier: I know this isn't new, but is there a nicer way of writing this without all the maps?

JohnWattsGDS: Well, as it is, I'm not entirely sure there is (barring methods to hide what are effectively a bunch of null checks) but it does open the wider question of should Optional<>s be used for anything but return values.

PR #368, opened 2019-09-05
Replace guava immutable collections with Java ones in tests

rhowe-gds:
Now that Java has {List,Set,Map}.of and friends, we don't need to use so much
guava.

This commit only touches tests.


rhowe-gds:
> Looks good - although we could also go the extra mile and get rid of Maps.newHashMap() too...

Extra miles are what future PRs are for :)

PR #370, opened 2019-09-06
BAU: Replace Google Optional with Java Util Optional

kerrr:
Replace `com.google.common.base.Optional` with built in `java.util.Optional`.


kerrr:
Hurm, makes changes to policy state objects, so is failing those tests. Potential ZDD failing change. need to think about this.

PR #374, opened 2019-09-09
If assertions are unsigned, we need to pass the original saml and als…

JohnWattsGDS:
If assertions are unsigned, we need to pass the original saml and also, symmetric encryption keys encrypted for the RP (MSA/VSP).

The main change is to make use of the changes in verify-saml-libs from [EID-1609](https://github.com/alphagov/verify-saml-libs/pull/86).  CountryAuthnResponseTranslatorService#translate returns an InboundResponseFromCountry that contains the original SAML request (for simplicity), a boolean stating whether or not the assertions are unsigned and conditional on that, re-encrypted symetric keys so the RP/VSP/MSA can read the encrypted assertions in the original SAML message.


hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/services/CountryAuthnResponseTranslatorServiceTest.java
```diff
@@ -114,5 +147,15 @@ public void shouldExtractAuthnStatementAssertionDetails() {
...
+
+        assertThat(result.areAssertionsUnsigned()).isFalse();
+        assertThat(result.getBase64Keys().size()).isEqualTo(0);
+    }
+
+    @Test public void shouldExtractAuthnStatementAssertionDetailsUnsignedAssertions() throws Exception {
+        setupUnisigned();
+        InboundResponseFromCountry result = service.translate(samlAuthnResponseTranslatorDto);
+        assertThat(result.areAssertionsUnsigned()).isTrue();
+        assertThat(result.getBase64Keys().size()).isEqualTo(1);
```
Wynndow:
```suggestion
        assertThat(result.getBase64Keys().size()).isEqualTo(List.of("key"));
```

PR #380, opened 2019-09-18
BAU Don't check detail for AuthenticationFailed status

Wynndow:
My knowledge of status's and the significance of details within them is
low is feedback on this is appreciated.

We've recently seen an increase in errors being thrown when an IDP
returns an AuthenticationFailed status. This is due to a [recent change
in the logic for mapping the status from the SAML to a Java object](https://github.com/alphagov/verify-hub/pull/369).

This status has a top-level status of Responder and a sub-status of
Authn_failed. The recent changes meant that we failed to match this
status when mapping if the IDP also included a status detail value (we
_only_ matched if there was no detail).

We added some logging to check if IDP's were including detail values for
this status and it turns out they were. [An example can be seen here.](https://kibana.logit.io/s/a7c5d5b3-a64d-4a60-961e-f9477bb11536/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:'2019-09-18T08:17:35.406Z',mode:absolute,to:'2019-09-18T08:17:35.519Z'))&_a=(columns:!(_source),index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',interval:auto,query:(language:lucene,query:'%22is%20not%20a%20valid%20sub-status%20when%20top-level%20status%20is%22%20OR%20%22Detail%20value%22'),sort:!('@timestamp',desc)))

As this status is the only status with this combination of top-level and
sub-status, we shouldn't care what the detail is and should match
regardless.

### DIFF
--- a/hub-saml/src/main/java/uk/gov/ida/saml/hub/transformers/inbound/SamlStatusToIdpIdaStatusMappingsFactory.java
+++ b/hub-saml/src/main/java/uk/gov/ida/saml/hub/transformers/inbound/SamlStatusToIdpIdaStatusMappingsFactory.java
@@ -15,27 +15,18 @@ import java.util.function.Predicate;
 
 public class SamlStatusToIdpIdaStatusMappingsFactory {
 
-    private static final Logger LOG = LoggerFactory.getLogger(SamlStatusToIdpIdaStatusMappingsFactory.class);
-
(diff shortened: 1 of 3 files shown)


Wynndow:
Have added a test on @kerrr 's advice

PR #385, opened 2019-09-27
BAU: Fix SAML Status mapping

kerrr:
We are getting some unexpected status mapping errors. This might be caused by our new stricter mapping code. I've relaxed the mapping to allowed for SAML status

`urn:oasis:names:tc:SAML:2.0:status:Responder` > `urn:oasis:names:tc:SAML:2.0:status:NoAuthnContext`

to have any details value.

Added logging so that any unmapped statuses are logged in detail.

Also simplified some mapping code for eidas that was over complex.


kerrr:
Alex, yes that serialises fine. I've added some more tests though.

PR #390, opened 2019-10-02
EID-1666 Handle unsigned assertions in a non-matching flow.

Wynndow:
Originally there were 20ish meandering commits to this PR. I've squashed them into two quite large commits, _hopefully_ making it easier to review. They're split by changes to policy and changes to saml-engine, with large commit messages (see below). It's worth noting that this flow includes *two* requests from the frontend to the hub, both which touch policy and saml-engine. I've got a sequence diagram I'm happy to share which may help make sense of what this PR is doing.

### Important
I'd really like opinions on the approach taken in saml-engine when transforming the response to a string. I've assumed that no services using the new non-matching flow will be using the legacy saml profile, and that they'll be using SHA256 for signing. This means saml-engine can skip calls to config. 

Commit messages are below.

### EID-1666 Update saml-engine to support unsigned assertions 
This commit does a couple of things.

When a response from a country is POSTed to the hub for the first time
it's details are stored in a state object by policy. Policy calls SAML
engine with a saml response DTO, and saml engine parses it and returns
an `InboundResponseFromCountry` dto. This inbound resonse is used to
create the state object in policy.

This updates saml-engine to check if the assertions in the saml are
unsigned, and if not, include a `CountrySignedResponseContainer` in the
`InboundResponseFromCountry` dto. This includes the original saml
response from the country and the encrypted assertion keys required by
the service to decrypt the assertions. These details are added to the
state in policy in another commit.

Later in the journey, the frontend calls the hub again, to retrieve the
details of the response store in the session state in policy. Policy
then calls saml-engine to create a signed saml Response to return to the
services. This commit updates saml-engine to support the case where a
countries assertions are unsigned.

saml-engine will return a wrapper Response, with the original country
saml-response as an attribute value, and the encrypted assertion keys as
another attribute value, within the assertion. These attributes are
newly created extensions in saml-lib (see a separate PR for these). Once
the Response has been generated, it's signed as normal, but the
assertions are not encrypted. This is because the sensitive information
is _already_ encrypted for the service, and so can be left as data in
the assertions.

The Response is then returned to Policy, and back to the frontend, and
finally to the user via a form. This approach of saml-in-saml allows us
to not break the api of the VSP, which will be receiving these
Responses. It's worth noting that this new flow is only the case for
non-matching journeys. Matching journeys will handle the unsigned
assertions in a separate flow which includes sending messages to the
services MSA.

### EID-1666 Update policy to support unsigned assertions
This updates policy to support unsigned assertion in a non-matching
journey.

When the hub initially receives a Response from a country, policy calls
saml-engine to parse the saml and receives an
`InboundResponseFromCountry` dto which includes details of the Response.
If the original Response included unsigned assertions, these details now
include the original Response from a country, as well as the encrypted
assertion keys required by the service. Policy will now store those
details in the `NonMatchingJourneySuccessState`. This allows them to be
retrieved later on.

Later on in the journey the frontend calls the hub again to ask it to
generate a signed response for the service. When this happens, policy
checks to see if the state of the request is
`NonMatchingJourneySuccessState`, and if it is, if it's assertions are
unsigned (by checking if the state includes a
`CountrySignedResponseContainer`).

If both of these are true, policy
calls a new endpoint on saml-engine specifically for dealing with
unsigned assertions from countries. Saml-engine will return a saml
Response with the origal saml Response from the country and the
encrypted assertion keys as attributes in it's assertions (see the
previous commit).

This Response is then returned all the way to the frontend and finally
to the user.

### EID-1666 Encrypt assertions in generated Response
The data within the assertions is already encrypted, however to remain
consistent with what we do elsewhere, such as sending the unsigned saml
response to the MSA, it makes sense to encrypt the new assertions too.

This has the benefit of allowing us to re-use the string transformers
which do the encryption for us, rather than pulling in a new one in
saml-lib.

To encrypt, we need to inject an `AssignableEntityToEncryptForLocator`
into the `RpAuthnResponseGeneratorService` and register the request ID
and entity to encrypt for.


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/factories/OutboundResponseFromHubToResponseTransformerFactory.java
```diff
@@ -49,4 +54,8 @@ public OutboundResponseFromHubToResponseTransformerFactory(OutboundLegacyRespons
         }
         return simpleProfileOutboundResponseFromHubToResponseTransformerProvider.get();
     }
+
+    public Function<AuthnResponseFromCountryContainerDto, String> getCountryTransformer() {
```
Wynndow: I've delved into the difference between a "legacy" response and a "samlProfile" response. The only difference is the status marshaller that they use. The "samlProfile" status marshaller will map a `NoMatchingServiceMatchFromHub` status slightly differently. In the non-matching flow being implemented here, the status is being hardcoded into the response as we already know we're in a success journey. See `SamlProfileTransactionIdaStatusMarshaller` and `TransactionIdaStatusMarshaller` for the implementation.

As for the SHA1/SHA256 switch, I've checked and all services have been upgraded to using SHA256. We could probably remove the extra call to config...

Finally, the only services that have `hubShouldSignResponseMessages` set to false in config are test-rp, and for some reason `dvsa-declaration` in integration (but not prod).

I think that means it's safe to skip the config calls in this flow and just return the new function here...

build.gradle
```diff
@@ -45,7 +45,7 @@ def dependencyVersions = [
         ida_test_utils:"2.0.0-46",
         opensaml:"$opensaml_version",
         dev_pki: '1.1.0-37',
-        saml_lib:"$opensaml_version-215"
+        saml_lib:"$opensaml_version-219"
```
idrop: It's worth taking 220 and then updating how we get and set values on CountrySamlResponse and EncryptedAssertionKeys

hub/policy/src/main/java/uk/gov/ida/hub/policy/controllogic/AuthnRequestFromTransactionHandler.java
```diff
@@ -122,4 +129,29 @@ public ResponseFromHub getErrorResponseFromHub(SessionId sessionId) {
...
+
+    public boolean isResponseFromCountryWithUnsignedAssertions(SessionId sessionId) {
+        try {
+            NonMatchingJourneySuccessStateController stateController = (NonMatchingJourneySuccessStateController)
+                    sessionRepository.getStateController(sessionId, ResponsePreparedState.class);
+            NonMatchingJourneySuccessState state = stateController.getState();
+            return state.getCountrySignedResponseContainer().isPresent();
+        } catch (ClassCastException e) {
+            return false;
+        }
```
idrop: Is this ClassCastException block necessary? We don't catch for it in the method above.

Wynndow: We do, because in a success flow there could be two different states at this point. One for a matching flow and one for non matching. If we're in the matching flow, the controller can't be cast to a `NonMatchingJourneySuccessStateController`, so we need this.

hub/policy/src/test/java/uk/gov/ida/hub/policy/domain/controller/EidasCountrySelectedStateControllerTest.java
```diff
@@ -63,6 +63,7 @@
     private static final String SELECTED_COUNTRY = STUB_COUNTRY_ONE;
     private static final String ANALYTICS_SESSION_ID = "some-session-id";
     private static final String JOURNEY_TYPE = "some-journey-type";
+    private static final Optional<CountrySignedResponseContainer> EMPTY_COUNTRY_SIGNED_RESPONSE = Optional.empty();
```
instantiator: The other params throughout this file just use `Optional.empty()` in place. I actually prefer what you've done here as it's much clearer what the parameter is for. Perhaps we should be consistent and do it for all the params?

hub/policy/src/test/java/uk/gov/ida/hub/policy/services/AuthnResponseFromCountryServiceTest.java
```diff
@@ -74,6 +74,7 @@
     private static final SamlAuthnResponseContainerDto SAML_AUTHN_RESPONSE_CONTAINER_DTO = aSamlAuthnResponseContainerDto().withSessionId(SESSION_ID).withPrincipalIPAddressAsSeenByHub("1.1.1.1").withAnalyticsSessionId(ANALYTICS_SESSION_ID).withJourneyType(JOURNEY_TYPE).build();
     private static final SamlAuthnResponseTranslatorDto SAML_AUTHN_RESPONSE_TRANSLATOR_DTO = SamlAuthnResponseTranslatorDtoBuilder.aSamlAuthnResponseTranslatorDto().build();
     private static final SamlAuthnResponseTranslatorDto NON_MATCHING_SAML_AUTHN_RESPONSE_TRANSLATOR_DTO = SamlAuthnResponseTranslatorDtoBuilder.aSamlAuthnResponseTranslatorDto().withMatchingServiceEntityId(TEST_RP).build();
+    private static final Optional<CountrySignedResponseContainer> EMPTY_UNSIGNED_ASSERTIONS = Optional.empty();
```
instantiator: Same comment as in `ErrorStateControllerTests`

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/resources/translators/RpAuthnResponseWrappingCountryResponseGeneratorResource.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Inject
+    public RpAuthnResponseWrappingCountryResponseGeneratorResource(RpAuthnResponseGeneratorService service) {
+        this.service = service;
+    }
+
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Timed
```
instantiator: Interesting - why do we time this?

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/services/RpAuthnResponseGeneratorServiceTest.java
```diff
@@ -0,0 +1,85 @@
...
+        );
+        generatorService = new RpAuthnResponseGeneratorService(
+                outboundResponseFromHubToResponseTransformerFactory,
+                HUB_ENTITY_ID,
+                new AssignableEntityToEncryptForLocator()
+        );
+    }
+
+    @Test
+    public void generateShouldReturnAnAuthnResponseFromHubContainerDtoFromAuthnResponseFromCountryContainerDto() {
```
instantiator: This is quite a long method name ("should return W response from X DTO from Y DTO from Z DTO"). I don't think there's much we can do about it though as it's all key to understanding what it does. Personal preference would be for a shorter (but still helpful) test method name, and a method comment explaining what it is doing. I know convention says not to do that, though. 🤷‍♂

hub-saml/src/main/java/uk/gov/ida/saml/hub/api/HubTransformersFactory.java
```diff
@@ -464,6 +486,14 @@ private OutboundResponseFromHubToSamlResponseTransformer getSamlProfileOutboundR
                 getEncryptedAssertionUnmarshaller());
     }
 
+    private OutboundAuthnResponseFromCountryContainerToSamlResponseTransformer getOutboundAuthnResponseFromCountryContainerToSamlResponseTransformer(String hubEntityId) {
+        return new OutboundAuthnResponseFromCountryContainerToSamlResponseTransformer(
+                new OpenSamlXmlObjectFactory(),
```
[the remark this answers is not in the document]
JohnWattsGDS: I can stick my nose in here and suggest that it's just the dependency injection pattern at work - this gives us the option to easily inject a mock during testing; if it's instantiated inside the class using mocks is no longer an option.

hub/policy/src/main/java/uk/gov/ida/hub/policy/controllogic/AuthnRequestFromTransactionHandler.java
galund: Or should there be a method `isNonMatchingFlowStateController` or something like that?

Wynndow: Have gone for the `instanceof` option. Might be a bit nicer now.

hub-saml/src/main/java/uk/gov/ida/saml/hub/api/HubTransformersFactory.java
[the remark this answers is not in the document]
Wynndow: As John says, I think it's better to pass the thing in so it's less tightly coupled to the `OpenSamlXmlObjectFactory` class. Having said that, I've looked at the test and we're not actually mocking the class, but doing it this way leaves it flexible for any future tests and is also consistent with how we set up other transformers in the `HubTransformerFactory`.

hub/policy/src/integration-test/java/uk/gov/ida/integrationtest/hub/policy/apprule/support/TestSessionResourceHelper.java
```diff
@@ -177,4 +183,35 @@ public static Response createSessionInEidasCountrySelectingState(SessionId sessi
...
+                null,
+                null,
+                false,
+                LevelOfAssurance.LEVEL_2,
+                false,
+                null,
+                nonMatchingJourneySuccessState.getTransactionSupportsEidas(),
+                nonMatchingJourneySuccessState.getEncryptedAssertions(),
+                nonMatchingJourneySuccessState.getCountrySignedResponseContainer().get()
+        );
```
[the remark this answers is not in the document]
Wynndow: That does sound good! I'm not going to implement here though, as it'd mean touching a lot of unrelated stuff and this PR is already too much.

hub/policy/src/test/java/uk/gov/ida/hub/policy/domain/controller/EidasCountrySelectedStateControllerTest.java
Wynndow: Agreed, have updated the rest of this file. I suspect there are other places that this could be done across the test suite, but that's a job for another day...

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/resources/translators/RpAuthnResponseWrappingCountryResponseGeneratorResource.java
Wynndow: Great question. Mostly because I copied the boilerplate from another endpoint. I suspect the timing goes into the access logs maybe? Someone may have been trying to profile saml-engine's performance at some point. ¯\_(ツ)_/¯

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/services/RpAuthnResponseGeneratorServiceTest.java
Wynndow: Yeah, it is a bit of a mouthful. It's actually "should return X Dto from Y Dto", it's just the Dto's are named "AuthnResponseFromHubContainerDto" and "AuthnResponseFromCountryContainerDto".

instantiator:
(Thanks for breaking out the commits Chris - I think it really does help to be able to see the various headliners separately.)

PR #396, opened 2019-10-16
EID-1721 Pass countrySignedResponseContainer to MSA for user account creation

Wynndow:
When the MSA doesn't find a user we enter the user account creation flow. In the case of unsigned assertions, we need to send the data from the `countrySignedResponseContainer` object as an assertion in the attribute query, rather than the original encrypted assertion.

Work has already been done to ensure that this object is available in the correct states, as well as the work to include the unsigned response assertion rather than the encrypted assertion (in `HubEidasAttributeQueryRequestToSamlAttributeQueryTransformer`).


instantiator:
If it helps with the 2nd review, the new tests from https://github.com/alphagov/verify-acceptance-tests/pull/73 pass when this PR is applied.

PR #398, opened 2019-10-16
BAU: Remove unnecessary predicates

kerrr:
The existing predicate logic was overly complex as it was written for a pre Java 8 world.

This first step is to remove the classes that implement Predicate<IdentityProviderConfig>, pulling the code from the test method into IdentityProviderConfig where needed. Where these predicates were used is then replaced with a method reference or lambda.

At this stage I have not tried to tackle refactoring the content of the predicate methods, as they are a little confusing. That will be simplified once I determine what they actually do.


hub/config/src/main/java/uk/gov/ida/hub/config/domain/IdentityProviderConfig.java
```diff
@@ -127,7 +131,16 @@ public Boolean isEnabledForSingleIdp() {
     }
 
     public boolean isOnboardingForTransactionEntity(String transactionEntity) {
-        return this.getOnboardingTransactionEntityIds().contains(transactionEntity);
+        boolean isOnboarding = !getOnboardingTransactionEntityIds().isEmpty();
```
galund: I don't feel like this line makes what's happening more clear :-)

kerrr: Really, this is just pulling around from one place to another. I agree it's not very clear, but for this stage I didn't what to change what the code does. The existing method wasn't being used, and was squatting the name that made it consistent with the other predicates. I plan on making another update that makes the code clearer more straightforward to read.

galund: Sure, but it'd still do the same thing with the existing implementation, I don't think you need the boolean at all?

kerrr: No, this does change the behaviour of this method. The new behaviour returns true if the list is empty.

As I say, I do think the intended behaviour could be much clearer, but I'm being cautious and not making any changes to the boolean logic being executed from the old predicates classes until I've fully understood it. Getting rid of the `Predicate` class cruft should help make it clearer what the actual intent is.

galund: Okay, in which case the function name now needs to change, because even now having read it 3 or 4 times I was confused :-)

kerrr: Removal of deprecated endpoints also removes this method.

hub/config/src/main/java/uk/gov/ida/hub/config/domain/IdentityProviderConfig.java
```diff
@@ -126,8 +130,12 @@ public Boolean isEnabledForSingleIdp() {
...
     }
 
-    public boolean isOnboardingForTransactionEntity(String transactionEntity) {
-        return this.getOnboardingTransactionEntityIds().contains(transactionEntity);
+    public boolean isOnboardingForTransactionEntityAtLoa(String transactionEntity, LevelOfAssurance levelOfAssurance){
+        boolean isOnboarding = levelOfAssurance != null ?
+                isOnboardingAtLoa(levelOfAssurance) :
+                isOnboardingAtAllLevels();
+
+        return !isOnboarding || getOnboardingTransactionEntityIds().contains(transactionEntity);
```
galund: The naming of this function (still?) seems wrong. If we return true if no services are onboarding, thencalling it `isOnboarding`-anything feels wrong? I feel like that part of the test should be outside the function, at the caller, if necessary – or we need a rename (but to what I've no idea).

kerrr: This is a method that replaces the deleted `OnboardingForTransactionEntityPredicate` class. That predicates logical expression as a boolean method is `isOnboardingForTransactionEntity`. That was previously already being used, and the predicate also used LOA. Given that, I think this is the correct name for the method at this stage. Because all the code around this is so confusing, I'm intentionally only making changes that involve moving the code around and not changing the wording. In the next step I will look at how this can all be made clearer and more easily understood. I think this is a reasonable stage for a PR. The alternative is to do the next step and ask reviewers to understand a much larger change, with more moving pieces.

PR #399, opened 2019-10-17
Update supported-msa-versions.yml for release 4.2.1-901

galund:
This change was supposed to be created by verify-release-automation, but it tried to push to master and failed :-(


hub/saml-soap-proxy/src/main/resources/supported-msa-versions.yml
```diff
@@ -5,3 +5,4 @@ versions:
   - 3.1.0-840
   - 4.0.0-871
   - 4.0.1-877
+  - 4.2.1-901
```
kerrr: Shouldn't 4.1.0-884 be in this list too?

galund: Yeah. We don't know whether we really care about this list or not. We also don't know why 4.1 never got added. Or whether we should be removing some older versions. Ho hum.

kerrr: I think (at some time?) it caused a sentry alert if an unsupported version was used.

PR #407, opened 2019-11-14
EID-1762 Move eIDAS hash logging logic to SAML lib

Wynndow:
We log hashes of eIDAS attributes in the hub as well as the proxy-node.
These are then compared in Splunk to ensure there's been no funny
business.

We'd like to consolidate all the logic for generating these hashes in
one place to ensure they're logged in the same way. This moves the logic
to saml-lib and pulls it in.


hub-saml/src/main/java/uk/gov/ida/saml/hub/transformers/inbound/providers/DecoratedSamlResponseToIdaResponseIssuedByIdpTransformer.java
```diff
@@ -1,8 +1,8 @@
 package uk.gov.ida.saml.hub.transformers.inbound.providers;
 
 import org.opensaml.saml.saml2.core.Response;
+import uk.gov.ida.eidas.logging.EidasAttributesLogger;
```
galund: I've just noticed that it feels like maybe this class is not in the right namespace if it's coming from samllibs?

Wynndow: What makes you say that - what would you suggest otherwise?

CharlesIC: I think the namespace is fine - we already have a bunch of namespaces in saml-libs that are....there

CharlesIC: Although admittedly `matchingserviceadapter` and `verifyserviceprovider` shouldn't be namespace names in a generic library. We can refactor this later, though

galund: Yeah, I think if it's in a shared library, then by definition we shouldn't have classes that are specific to either eIDAS, or to the MSA, or the VSP – and the namespaces would reflect that.

CharlesIC: Totally agree, George 👍

PR #413, opened 2019-11-23
6n6H8nhi -Make cache expiry for certs configurable

alan-gds:
https://trello.com/c/6n6H8nhi/776-make-cache-expiry-for-certs-configurable-in-saml-engine-saml-soap-proxy-and-saml-proxy

Three of the Hub services (saml-proxy, saml-soap-proxy and saml-engine) read federation certs from the config service, and store them in a cache.  The cache was formerly hard-coded to have an expiry time of five minutes.  The self-service cert rotation app needs to wait for this cache to expire in some circumstances, so we want to reduce it to a shorter interval.  We'd prefer for it to be configurable rather than hard-coded (in particular, this should help facilitate testing).

The cache expiry time is set to one minute in the config files.

https://trello.com/c/6n6H8nhi/776-make-cache-expiry-for-certs-configurable-in-saml-engine-saml-soap-proxy-and-saml-proxy

### DIFF
--- a/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineConfiguration.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineConfiguration.java
@@ -103,7 +103,7 @@ public class SamlEngineConfiguration extends Configuration implements RestfulCli
     @Valid
     @NotNull
     @JsonProperty
-    protected Duration certificatesConfigCacheExpiry;
+    protected Duration certificatesConfigCacheExpiry = Duration.minutes(5);
 
     @Valid
     @NotNull
--- a/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineConfiguration.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineConfiguration.java
@@ -103,6 +103,11 @@ public class SamlEngineConfiguration extends Configuration implements RestfulCli
     @Valid
     @NotNull
     @JsonProperty
+    protected Duration certificatesConfigCacheExpiry;
+
+    @Valid
+    @NotNull
+    @JsonProperty
     protected ClientTrustStoreConfiguration rpTrustStoreConfiguration;
 
     public SamlConfiguration getSamlConfiguration() {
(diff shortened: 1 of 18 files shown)


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineConfiguration.java
```diff
@@ -100,6 +100,11 @@ protected SamlEngineConfiguration() {}
     @JsonProperty
     protected URI configUri;
 
+    @Valid
+    @NotNull
+    @JsonProperty
+    protected Duration certificatesConfigCacheExpiry;
```
kerrr: Might be nice to have a default value here.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
```diff
@@ -495,6 +495,13 @@ private URI configUri(SamlEngineConfiguration configurations) {
         return configurations.getConfigUri();
     }
 
+    @Provides
+    @Singleton
+    @Config
+    private long certificatesConfigCacheExpiryInSeconds(SamlEngineConfiguration configuration) {
+        return configuration.getCertificatesConfigCacheExpiry().toSeconds();
+    }
```
kerrr: Any reason why we don't just use the duration and transform it when we need it?

kerrr:
Oh, you should probably clean up those Codacy gripes. Ping me and I'll re-approve.

jakubmiarka:
> Oh, you should probably clean up those Codacy gripes. Ping me and I'll re-approve.

Alan said: "_Codacy is flipping out about some perfectly reasonable DI code and the fact that the configuration classes have a constructor declaration above the field declarations; all of these are me just following the conventions we're already using and propagating existing (non-)issues._"

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineConfiguration.java
jakubmiarka: agree, we'll not need this PR then https://github.com/alphagov/ida-hub-acceptance-tests/pull/159

jakubmiarka: made the change

PR #414, opened 2019-12-05
TECH-DEBT: Upgrade test libs, dependencies and Gradle to latest versions

CharlesIC:
Recommend reviewing this PR by commits.
This PR is pulling changes from alphagov/verify-saml-libs#107

- Upgrade Gradle and test libraries to latest versions
Explicitly define test dependencies (JUnit, Mockito, AssertJ) in the Gradle file as these are no longer exposed in the new version of the test utils library.

- Move logging of eIDAS attributes hash from `DecoratedSamlResponseToIdaResponseIssuedByIdpTransformer` to `IdpAuthnResponseTranslatorService`.
This makes more sense as the service has all the information it needs to decide when it should log the hash and to call the logger. This also removes the clunky pattern of setting the logger conditionally in the transformer and avoids a potential bug where the transformer would log more often than necessary.

- Compliance with Gradle 6
  - Remove deprecated configuration options - replace `compile` with `implementation` and `testCompile` with `testImplementation`
  - Apply the new `java-library` plugin and library projects and use the `api` configuration to expose dependencies used by consumers
  - Remove redundant test libraries

- Update travis curl script used to download coverage reporter jar

- Stop using deprecated methods
  - Don't use deprecated Mockito methods. Switch to new ones.
  - Use AssertJ fluent assertions everywhere.
We mostly use AssertJ assertions so replace usages of JUnit and Hamcrest style assertions with AssertJ.
  - Use generics to avoid unchecked casts and using the raw version of parametrised classes


hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/exceptions/ExceptionAuditor.java
```diff
@@ -25,7 +25,7 @@ public boolean auditException(ApplicationException exception, Optional<SessionId
         if (!isAudited && exception.requiresAuditing()) {
             EventDetails eventDetails = new EventDetails(
                     downstream_uri,
-                    exception.getUri().or(URI.create("uri-not-present")).toASCIIString());
```
JohnWattsGDS: Why the change to `orElse`?  Doesn't `or` or `orElseGet` only create the URI (invoke the supplier) if the Optional is empty? Whereas `orElse` will always create the URI. I know I like to say don't prematurely optimise but this would seem to be a case where it's better stick with a thing that does as little as possible?

CharlesIC: This is because saml-libs was switched from Guava Optionals to Java Optionals. The equivalent of Guava's `or` is `orElse`

JohnWattsGDS: Oh, I missed the change of import in the commit - it looked like it had been the native Optionals for a while ...

hub/saml-proxy/src/main/java/uk/gov/ida/hub/samlproxy/exceptions/ExceptionAuditor.java
```diff
@@ -25,7 +25,7 @@ public boolean auditException(ApplicationException exception, Optional<SessionId
...
+                    exception.getUri().orElse(URI.create("uri-not-present")).toASCIIString());
```
JohnWattsGDS:
```suggestion
                    exception.getUri().orElseGet(() -> URI.create("uri-not-present")).toASCIIString());
```

JohnWattsGDS: That's the fancy-pants approach anyway ...

hub-saml/src/test/java/uk/gov/ida/saml/hub/domain/LevelOfAssuranceTest.java
```diff
@@ -2,19 +2,18 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class LevelOfAssuranceTest {
 
     @Test
     public void checkOrdering() {
-        assertTrue(LevelOfAssurance.HIGH.compareTo(LevelOfAssurance.SUBSTANTIAL) > 0);
-        assertTrue(LevelOfAssurance.HIGH.compareTo(LevelOfAssurance.LOW) > 0);
-        assertTrue(LevelOfAssurance.SUBSTANTIAL.compareTo(LevelOfAssurance.LOW) > 0);
+        assertThat(LevelOfAssurance.HIGH.compareTo(LevelOfAssurance.SUBSTANTIAL) > 0);
```
JohnWattsGDS:
```suggestion
        assertThat(LevelOfAssurance.HIGH.compareTo(LevelOfAssurance.SUBSTANTIAL)).isGreaterThan(0);
```

JohnWattsGDS: How do you feel about using this approach?  It requires less cognitive processing and is consistent with the other 'new' methods.

CharlesIC: Fluent assertions using 'proper' checks like this gives better error message. So if the test fails instead of seeing something like "Expected false to be true" you will see "Expected x to be greater than y"

hub-saml/src/main/java/uk/gov/ida/saml/hub/validators/response/idp/components/EncryptedResponseFromIdpValidator.java
```diff
@@ -27,7 +27,7 @@
 import static uk.gov.ida.saml.core.errors.SamlTransformationErrorFactory.unexpectedNumberOfAssertions;
 import static uk.gov.ida.saml.security.validators.signature.SamlSignatureUtil.isSignaturePresent;
 
-public class EncryptedResponseFromIdpValidator<T extends Enum> {
+public class EncryptedResponseFromIdpValidator<T extends Enum<?>> {
```
adityapahuja: Is there any benefit of using `<T extends Enum<?>>` instead of `<T extends Enum<T>>`?

hub-saml/src/test/java/uk/gov/ida/saml/hub/domain/LevelOfAssuranceTest.java
adityapahuja: I swapped both level of assurances on this line and ran the test again. It did not fail. 

`assertThat(LevelOfAssurance.SUBSTANTIAL.compareTo(LevelOfAssurance.HIGH) > 0);`

`assertThat()` didn't check compareTo() result. We should use John's suggested change.

hub/saml-soap-proxy/src/test/java/uk/gov/ida/hub/samlsoapproxy/healthcheck/HealthCheckDataTest.java
```diff
@@ -3,52 +3,50 @@
...
     @Test
     public void shouldReturnHealthCheckDataWithVersionWhenExtractedFromStringThatContainsVersionData() {
         String expectedVersion = "1234";
         HealthCheckData testData = HealthCheckData.extractFrom("test-id-version-" + expectedVersion);
-        assertFalse(testData.getEidasEnabled().isPresent());
-        assertFalse(testData.getShouldSignWithSha1().isPresent());
-        assertTrue(testData.getVersion().get().contains(expectedVersion));
+        assertThat(testData.getEidasEnabled().isPresent()).isFalse();
+        assertThat(testData.getShouldSignWithSha1().isPresent()).isFalse();
+        assertThat(testData.getVersion().get().contains(expectedVersion));
```
adityapahuja: Can we replace `assertThat(testData.getVersion().get().contains(expectedVersion));` with `assertThat(testData.getVersion().get().contains(expectedVersion)).isTrue();`?

hub/saml-soap-proxy/src/test/java/uk/gov/ida/hub/samlsoapproxy/healthcheck/HealthCheckDataTest.java
```diff
@@ -3,52 +3,50 @@
...
     @Test
     public void shouldReturnHealthCheckDataWithVersionAndEidasWhenExtractedFromStringThatContainsVersionDataAndEidas() {
         String expectedVersion = "1234";
         String expectedEidasEnabled = "true";
         HealthCheckData testData = HealthCheckData.extractFrom("test-id-version-" + expectedVersion + "-eidasenabled-" + expectedEidasEnabled);
-        assertTrue(testData.getVersion().get().contains(expectedVersion));
-        assertTrue(testData.getEidasEnabled().get().contains(expectedEidasEnabled));
-        assertFalse(testData.getShouldSignWithSha1().isPresent());
+        assertThat(testData.getVersion().get().contains(expectedVersion));
+        assertThat(testData.getEidasEnabled().get().contains(expectedEidasEnabled));
```
adityapahuja: Can we replace them with the following code?
```
assertThat(testData.getVersion().get().contains(expectedVersion)).isTrue();
assertThat(testData.getEidasEnabled().get().contains(expectedEidasEnabled)).isTrue();
```

hub/saml-soap-proxy/src/test/java/uk/gov/ida/hub/samlsoapproxy/healthcheck/HealthCheckDataTest.java
```diff
@@ -59,12 +57,12 @@ public void shouldReturnFullHealthCheckDataWhenExtractedFromStringContainingAllD
...
-                "-shouldsignwithsha1-" + expectedShouldSignWithSha1
+                        "-eidasenabled-" + expectedEidasEnabled +
+                        "-shouldsignwithsha1-" + expectedShouldSignWithSha1
         );
-        assertTrue(testData.getVersion().get().contains(expectedVersion));
-        assertTrue(testData.getEidasEnabled().get().contains(expectedEidasEnabled));
-        assertTrue(testData.getShouldSignWithSha1().get().contains(expectedShouldSignWithSha1));
+        assertThat(testData.getVersion().get().contains(expectedVersion));
+        assertThat(testData.getEidasEnabled().get().contains(expectedEidasEnabled));
+        assertThat(testData.getShouldSignWithSha1().get().contains(expectedShouldSignWithSha1));
```
adityapahuja: Can we replace them with the following code?

```
assertThat(testData.getVersion().get().contains(expectedVersion)).isTrue();
assertThat(testData.getEidasEnabled().get().contains(expectedEidasEnabled)).isTrue();
assertThat(testData.getShouldSignWithSha1().get().contains(expectedShouldSignWithSha1)).isTrue();
```

hub/saml-soap-proxy/src/test/java/uk/gov/ida/hub/samlsoapproxy/healthcheck/HealthCheckDataTest.java
```diff
@@ -73,11 +71,11 @@ public void shouldReturnHealthCheckDataWithVersionWithShaNoiseWhenExtractedFromS
...
         HealthCheckData testData = HealthCheckData.extractFrom(
-                "test-id-version-" + expectedVersion+
-                "-shouldsignwithsha1-" + expectedShouldSignWithSha1);
+                "test-id-version-" + expectedVersion +
+                        "-shouldsignwithsha1-" + expectedShouldSignWithSha1);
 
-        assertTrue(testData.getVersion().get().contains(expectedVersion+"-shouldsignwithsha1-" + expectedShouldSignWithSha1));
-        assertFalse(testData.getEidasEnabled().isPresent());
-        assertFalse(testData.getShouldSignWithSha1().isPresent());
+        assertThat(testData.getVersion().get().contains(expectedVersion + "-shouldsignwithsha1-" + expectedShouldSignWithSha1));
```
adityapahuja: Can we replace this line with the following code?
```
assertThat(testData.getVersion().get().contains(expectedVersion + "-shouldsignwithsha1-" + expectedShouldSignWithSha1)).isTrue();
```

hub-saml/src/test/java/uk/gov/ida/saml/hub/domain/LevelOfAssuranceTest.java
CharlesIC: Accepted change

hub-saml/src/main/java/uk/gov/ida/saml/hub/validators/response/idp/components/EncryptedResponseFromIdpValidator.java
CharlesIC: Yes - this avoids unchecked cast exceptions by using Java generics "properly"

hub/saml-soap-proxy/src/test/java/uk/gov/ida/hub/samlsoapproxy/healthcheck/HealthCheckDataTest.java
CharlesIC: It doesn't make much of a difference  - the fluent assertion message wouldn't contain any more detail so I think it's not necessary.

adityapahuja: It does make difference. I tried the original code to make it fail by making the version different. It didn't fail.

I added `.isTrue()` and ran the following code again. It failed. This showed that the test was working because `.isTrue()` was added.

```
    @Test
    public void shouldReturnHealthCheckDataWithVersionWhenExtractedFromStringThatContainsVersionData() {
        String expectedVersion = "1234";
        HealthCheckData testData = HealthCheckData.extractFrom("test-id-version-" + "1235");
        assertThat(testData.getEidasEnabled().isPresent()).isFalse();
        assertThat(testData.getShouldSignWithSha1().isPresent()).isFalse();
        assertThat(testData.getVersion().get().contains(expectedVersion)).isTrue();
    }
```

adityapahuja: We should use `assertThat(testData.getVersion().get()).contains(expectedVersion);` instead of `assertThat(testData.getVersion().get().contains(expectedVersion));` It gives better failure message.

hub/saml-soap-proxy/src/test/java/uk/gov/ida/hub/samlsoapproxy/healthcheck/HealthCheckDataTest.java
```diff
@@ -3,52 +3,50 @@
...
 
 public class HealthCheckDataTest {
 
     @Test
     public void shouldReturnEmptyHealthCheckDataWhenExtractedFromNullString() {
         HealthCheckData testData = HealthCheckData.extractFrom("test-id");
-        assertFalse(testData.getVersion().isPresent());
-        assertFalse(testData.getEidasEnabled().isPresent());
-        assertFalse(testData.getShouldSignWithSha1().isPresent());
+        assertThat(testData.getVersion().isPresent()).isFalse();
```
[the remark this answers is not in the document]
CharlesIC: Whaaat?! I didn't know that was an option - but I don't want to dismiss the two approvals now so will make a note to add it next time 👍

hub/saml-soap-proxy/src/test/java/uk/gov/ida/hub/samlsoapproxy/healthcheck/HealthCheckDataTest.java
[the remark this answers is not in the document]
CharlesIC: It _does_ actually add more detail because the first `contains` is from `java.lang.String` and it will only return true/false to the test framework but the second `contains` is from AssertJ and it will print a more detail message 👍

PR #421, opened 2020-02-06
BAU: Hide sentry null exceptions due to certificates managed by self service tool

jkunle:
This is due to certificates managed by the Cert Admin tool
we will turn this back from warning to error once addressed by
https://govukverify.atlassian.net/browse/HUB-457


galund:
Are we sure this is the correct line to be changing from error to warning? I would have thought it was the other one.

jkunle:
https://sentry.tools.signin.service.gov.uk/verify/prod-hub/issues/4098/

PR #422, opened 2020-02-07
Remove infinispan

philandstuff:
Policy and saml-engine used to use infinispan as a session store.  We
migrated to use Redis instead, but for a transition period we had a
lot of conditional code which allowed the use of either backend.

The time has come to get rid of that code.

Of note:

 - stop using the old PolicyAppRule (which runs infinispan) and use
   PolicyAppRuleWithRedis instead (now renamed to PolicyAppRule)
 - remove StateDeserializationTest, because Redis uses JSON
   serialization instead of old skool java serialization


jakubmiarka:
Does it change anything when running locally? I don't think it uses Redis locally...

philandstuff:
@jakubmiarka probably, I didn't look at that. Could I get some help with this? I don't know how local startup works...

jakubmiarka:
> @jakubmiarka probably, I didn't look at that. Could I get some help with this? I don't know how local startup works...

@philandstuff we don't use verify-local-startup ... it's set of startup.sh scripts... to run the whole hub locally we use the ida-hub-acceptance test... there's a bunch of config which it uses, so the infinityspan should be replaced with redis there too? https://github.com/alphagov/ida-hub-acceptance-tests/blob/master/configuration/hub/saml-engine.yml#L33

and then add a docker for the redis or however we recommend spinning redis locally? somewhere here https://github.com/alphagov/ida-hub-acceptance-tests/blob/master/scripts/hub-apps-startup.sh

galund:
Oh, this is interesting, I think this means the folk-memory "reasons why ZDD tests tends to fail" things I thought I knew are wrong :-) 🎉

philandstuff:
Raised https://github.com/alphagov/ida-hub-acceptance-tests/pull/164 to try to fix that? don't know if it works?

configuration/local/policy.yml
```diff
@@ -15,15 +15,6 @@ logging:
     - type: console
 
 
-sessionStore:
```
jakubmiarka: should this be replaced with redis config instead? even as a placeholder maybe?

philandstuff:
@jakubmiarka nice spot, pushed an update to add sensible defaults to local config

PR #431, opened 2020-03-11
HUB-520 crash config service unable to load cert

jkunle:
https://govukverify.atlassian.net/browse/HUB-520


hub/config/src/main/java/uk/gov/ida/hub/config/data/ManagedEntityConfigRepository.java
```diff
@@ -36,27 +39,61 @@ public boolean has(String entityId){
...
+    private List<String> getSignatureVerificationCertificates(T local, RemoteComponentConfig remote) {
+        try {
+            return remote.getSignatureVerificationCertificates();
+        }
+        catch (NullPointerException e){
+            LOG.warn("Remote config signing certificates missing for {}",
+                    Optional.ofNullable(local.getSignatureVerificationCertificates()
+                            .stream().map(Certificate::getSubject)
+                            .reduce("", String::join))
+                            .orElseGet(local::getEntityId));
```
JohnWattsGDS: How does this work if the local config (which I presume to be the cert in the config file in `verify-hub-federation-config`) is missing?  The way I read this is that it falls back on the `v-h-f-c` cert if the remote one is missing.

jkunle: If it is missing then `.orElseGet(local::getEntityId));` it uses the `entity_id`

hub/config/src/main/java/uk/gov/ida/hub/config/data/ManagedEntityConfigRepository.java
```diff
@@ -36,27 +39,61 @@ public boolean has(String entityId){
...
+        try {
+            return remote.getEncryptionCertificate();
+        }
+        catch (NullPointerException e){
+            LOG.warn("Remote config encryption certificate missing for {}",
+                    local.getEncryptionCertificate().getX509Certificate()
+                            .map(X509Certificate::getSubjectDN)
+                            .map(Principal::getName)
+                            .orElseGet(local::getEntityId));
+            throw new NoCertificateFoundException();
```
CharlesIC: It seems weird to log a warning and then throw an error - it would be more consistent to create the `NoCertificateFoundException` with the message in the constructor instead of logging it - this way it will be correctly logged as "ERROR" and will show in the stack trace with the exception.

hub/config/src/test/java/uk/gov/ida/hub/config/data/ManagedEntityConfigRepositoryTest.java
```diff
@@ -110,5 +110,36 @@ public void getReturnsOverriddenConfigWhenRemoteFoundAndEnabled() {
         assertThat(result.get().getEncryptionCertificate().getBase64Encoded().get()).isEqualTo(REMOTE_CERT);
         assertThat(result.get().getEncryptionCertificate().getCertificateOrigin()).isEqualTo(CertificateOrigin.SELFSERVICE);
     }
+    
+    @Test(expected = NoCertificateFoundException.class)
```
CharlesIC: I'm confused by this test - if we're expecting the `get` to throw then why do we have the asserts? If `get` throws the test will never get that far down? 'd prefer to use `assertThatThrownBy(() ​-> ...)` instead of `(expected = ...)` so it's clear which line we're expecting to throw

hub/config/src/test/java/uk/gov/ida/hub/config/data/ManagedEntityConfigRepositoryTest.java
```diff
@@ -110,5 +110,36 @@ public void getReturnsOverriddenConfigWhenRemoteFoundAndEnabled() {
...
+                .withSelfService(true)
+                .build();
+        
+        when(localConfigRepository.getData(REMOTE_ENABLED_ENTITY_ID_2)).thenReturn(Optional.of(remoteEnabledTransactionWithoutEncryption));
+        Optional<TransactionConfig> result = configRepo.get(REMOTE_ENABLED_ENTITY_ID_2);
+
+        assertThat(result.get().getEntityId()).isEqualTo(REMOTE_ENABLED_ENTITY_ID_2);
+        assertThat(result.get().getEncryptionCertificate().getBase64Encoded()).isPresent();
+        assertThat(result.get().getEncryptionCertificate().getCertificateOrigin()).isEqualTo(CertificateOrigin.SELFSERVICE);
+    }
```
CharlesIC:
```suggestion
    }
    
```

hub/config/src/main/java/uk/gov/ida/hub/config/data/ManagedEntityConfigRepository.java
jkunle: I agree thanks

PR #434, opened 2020-03-20
HUB-545: Fix logic for disconnecting IDPs for registration

CharlesIC:
This addresses the scenario in which:
  - An IDP has the `provideRegistrationUntil` flag to set to a `cut-off time`
  - A user starts a registration journey and clicks the button to select the IDP on the picker page before the `cut-off time` ("registration authentication request")
  - The user then completes the IDP journey and successfully registers
  - On their return to the Hub they see the "Something wrong page" ("registration authentication response")

The user sees the error page on their return because at that point the IDP is no longer available for registration according to the config. To solve this, we need to accept registration responses for up to the duration of the Hub session (90 mins) after the cut-off time.

These changes make the distinction in the code between making an authentication request and receiving a response for determining the list of valid IDPs. For the request case the logic is the same - show the IDP if cut-off time is in the future. For the response case, the logic is changed to accept the IDP if `cut-off time + session duration` is in the future.


configuration/config.yml
```diff
@@ -8,6 +8,7 @@ server:
       - type: access-logstash-console
 serviceInfo:
   name: config
+userHubSessionDuration: 90m
```
galund: Surely this is in config somewhere already?

CharlesIC: Nope 🤷‍♂ 
It's already hardcoded in two other config yamls - making this a global config value is not in scope of this PR (I would do it anyway but this has to go out _today_ so not the best time)

galund: Understood - if you've found where those places are it might be worth a TODO somewhere to tie them all together.

hub/config/src/integration-test/java/uk/gov/ida/integrationtest/hub/config/apprule/IdentityProviderResourceIntegrationTest.java
```diff
@@ -206,27 +219,44 @@ public void getIdpConfigData_returnsNotFoundIfIdpDoesNotExist() {
...
+    }
+
     @Test
     public void getEnabledIdentityProviderEntityIdsForSignIn_returnsOkAndIdps() {
         String entityId = "not-test-rp";
         Response response = getIdpListForLoA(entityId, LevelOfAssurance.LEVEL_1, Urls.ConfigUrls.ENABLED_ID_PROVIDERS_FOR_SIGN_IN_RESOURCE);
         assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
 
-        List<String> providerEntityIds = response.readEntity(new GenericType<List<String>>(){});
+        List<String> providerEntityIds = response.readEntity(new GenericType<List<String>>() {
```
galund: If these formatting changes really needed to happen, I'd have preferred them in a separate commit, but no big deal. It looks neater on one line, FWIW.

configuration/config.yml
CharlesIC: I might push another PR after all this is done - as we'll need a new endpoint on the config service

hub/config/src/integration-test/java/uk/gov/ida/integrationtest/hub/config/apprule/IdentityProviderResourceIntegrationTest.java
CharlesIC: Fair, I was being lazy....

PR #443, opened 2020-04-08
[DO NOT MERGE] Event recording modification in anticipation of modifications to LOAs for some RPs

instantiator:
**This change should not affect current behaviour.**

This change amends `Policy` to allow it to make a better decision about which of the LoAs in an RPs config should be reported as **minimum**, and which as **required** (in Verify, believed to mean 'preferred'). This is not expected to change the behaviour for any current RPs.

It removes the assumption that the list of LoAs is ordered, in anticipation of changes being planned to permit some services to make a request with a different ordering in order to satisfy different behaviours.

The key changes are in:

* `//hub/policy/domain/controller/EidasCountrySelectedStateController.java`
* `//hub/policy/domain/controller/IdpSelectedStateController.java`

In each, instead of assigning the first level in the list to **minimum**, and the last to **required**, the lowest and highest are assigned respectively.

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/LevelOfAssurance.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/LevelOfAssurance.java
@@ -1,5 +1,7 @@
 package uk.gov.ida.hub.policy.domain;
 
+import java.util.List;
+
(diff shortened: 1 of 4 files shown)


instantiator:
Planning complete and we're working on a different PR for the hub changes.

PR #444, opened 2020-04-09
Allow an RP to have config where LOA2 is followed by LOA1

JHjava:
- We now have an RP which has LOA2 first in config followed by LOA1. This will change the order in which the LOAs appear in the AuthnRequest so that LOA2 is first in the list. This is to show that the RP prefers LOA2 but will accept LOA1 if a user fails to get LOA2.
- Update the config validator to allow this ordering of LOAs in an RP's transaction config.

### DIFF
--- a/hub/config/src/test/java/uk/gov/ida/hub/config/data/LevelsOfAssuranceConfigValidatorTest.java
+++ b/hub/config/src/test/java/uk/gov/ida/hub/config/data/LevelsOfAssuranceConfigValidatorTest.java
@@ -83,6 +83,16 @@ public class LevelsOfAssuranceConfigValidatorTest {
     }
 
     @Test
+    public void shouldAllowLOA2AndLOA1TransactionConfiguration() {
+        Set<TransactionConfig> transactionsConfig = Set.of(loa2And1Transaction);
+        try {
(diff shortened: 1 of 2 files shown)


hub/config/src/test/java/uk/gov/ida/hub/config/data/LevelsOfAssuranceConfigValidatorTest.java
```diff
@@ -82,6 +82,16 @@ public void shouldAllowLOA1TransactionConfiguration() {
         }
     }
 
+    @Test
+    public void shouldAllowLOA2AndLOA1TransactionConfiguration() {
+        Set<TransactionConfig> transactionsConfig = Set.of(loa2And1Transaction);
```
instantiator, line 87: I can't see where `loa2And1Transaction` is defined.

instantiator: Expanded the class - now I can see it. Has it always been there?

JHjava: Yes, because previously it was used for a test to ensure that this order of LOAs could not be used.

PR #445, opened 2020-04-22
Only send a [LOA2,LOA1] ordering of LOAs to IDPs that are enabled for registration.

JHjava:
We want IDPs which are enabled for receiving registration requests to be able to handle an Authn Request where the ordering of LOAs in the request is [LOA2, LOA1]. This is so that these IDPs will first attempt to achieve LOA2 for the User but failing that will try for LOA1. When a User is on the IDP in a sign-in journey they can switch to a registration journey, so it’s important that this ordering is used in both instances to ensure the IDP behaviour is consistent.

When a Service is configured with the [LOA2, LOA1] ordering of LOAs and they are sending a request to an IDP which is only configured for sign-in, then we want the request to contain just [LOA2]. This will be sent as a LOA2 request meaning a user who obtained LOA2 on registering can use the service. This will give the service the existing behaviour when sending requests to IDPs that are configured for only sign-in.

The change implemented here entails the following -

- Call config service to see if the IDP is enabled for registration.
- If it isn’t and the RP has the ordering of LOAs [LOA2, LOA1] then just use [LOA2].
- Add new tests to test this behaviour.

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/IdpSelector.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/IdpSelector.java
@@ -8,6 +8,7 @@ import uk.gov.ida.hub.policy.domain.state.IdpSelectingState;
 import uk.gov.ida.hub.policy.proxy.IdentityProvidersConfigProxy;
 import uk.gov.ida.hub.policy.proxy.TransactionsConfigProxy;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -34,7 +35,17 @@ public class IdpSelector {
 
         IdpConfigDto idpConfig = identityProvidersConfigProxy.getIdpConfig(idpEntityId);
         final List<LevelOfAssurance> idpLevelsOfAssurance = idpConfig.getSupportedLevelsOfAssurance();
-        List<LevelOfAssurance> levelsOfAssuranceForTransactionSupportedByIdp = levelsOfAssuranceForTransaction.stream().filter(idpLevelsOfAssurance::contains).collect(Collectors.toList());
+
+        boolean idpEnabledForRegistration = identityProvidersConfigProxy.isIDPEnabledForRegistration(
+                idpEntityId,
+                state.getRequestIssuerEntityId(),
+                requestedLoa);
+
+        List<LevelOfAssurance> levelsOfAssuranceForTransactionSupportedByIdp =
+                getLevelsOfAssuranceForTransactionSupportedByIdp(
+                        idpEnabledForRegistration,
+                        idpLevelsOfAssurance,
+                        levelsOfAssuranceForTransaction);
 
         return new IdpSelectedState(
                 state.getRequestId(),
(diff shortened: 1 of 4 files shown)


hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/IdpSelector.java
```diff
@@ -54,6 +64,25 @@ public static IdpSelectedState buildIdpSelectedState(IdpSelectingState state,
         );
     }
 
+    private static List<LevelOfAssurance> getLevelsOfAssuranceForTransactionSupportedByIdp(boolean idpEnabledForRegistration,
+                                                                                           List<LevelOfAssurance> idpLevelsOfAssurance,
+                                                                                           List<LevelOfAssurance> levelsOfAssuranceForTransaction) {
+        List<LevelOfAssurance> levelsOfAssuranceForTransactionSupportedByIdp;
+
+        if (!idpEnabledForRegistration
```
JohnWattsGDS, line 72: I guess this is written for scenarios where there are LOAs that aren't just 1 or 2 but, as that's all we have right now and on the basis of YAGNI, you could (unless I'm mistaken which is likely enough) simplify this to:

`if(!idpEnabledForRegistration && levelsOfAssuranceForTransaction.size() == 2) {
    levelsOfAssuranceForTransaction = List.of(LevelOfAssurance.LEVEL_1, LevelOfAssurance.LEVEL_2)
}`

and maybe even (although I don't think simplify is necessarily the correct term anymore):

`if(!idpEnabledForRegistration) {
    Collections.sort(levelsOfAssuranceForTransaction) // I'm assuming without checking the sort is ascending by default.
}`

JHjava: That makes a lot more sense, John. Good suggestion. I'm thinking to use your first suggestions as it is more explicit.

JHjava: I guess as it is now makes it clearer in stating the scenario in which the LOAs need to be switched? I'm thinking what would make more sense to someone reading this down the line 🤷‍♂️ But setting it using List.of is better so I'm going to change that and leave the rest for more feedback.

JHjava: I've had to update this based on newly discovered requirements. So in the case where registration is not enabled for the service and they have the LOA mapping of [LOA2, LOA1] then we now only want to send them LOA2.

hub/policy/src/main/java/uk/gov/ida/hub/policy/domain/controller/IdpSelector.java
```diff
@@ -54,6 +65,24 @@ public static IdpSelectedState buildIdpSelectedState(IdpSelectingState state,
...
+                && levelsOfAssuranceForTransaction.size() == 2
+                && levelsOfAssuranceForTransaction.indexOf(LevelOfAssurance.LEVEL_2) == 0
+                && levelsOfAssuranceForTransaction.indexOf(LevelOfAssurance.LEVEL_1) == 1) {
+            levelsOfAssuranceForTransaction = Collections.singletonList(LevelOfAssurance.LEVEL_2);
```
instantiator, line 77: As we're hard coding this, lets add a small comment that refers to the ADR.

JHjava: Done.

PR #451, opened 2020-05-29
Make sure the journey type and analytic session ID exist when we log it

JHjava:
**What**
- Add the analytic session id and journey type to the saml proxy dto so it is able to receive them from the frontend before logging them in policy.
- Make the necessary changes to the unit tests.

**Why**
- This is half of the change to ensure we log the analytic session id and journey type when we receive a response from an idp. The other half will be in verify-frontend.
- Currently we log the journey type and analytic session id when we receive a response from a country or IDP. However these attributes are always null. This is because the frontend should be adding these attributes to the response before passing it onto saml proxy.

This will work for IDPs. I'm not sure if we need to log either of these attributes for countries on the response. We're not currently logging an analytic session ID for countries on the request either. If we don't need to do this then we can stop trying to log it for countries and code can be removed.

### DIFF
--- a/hub/policy/src/test/java/uk/gov/ida/hub/policy/logging/HubEventLoggerTest.java
+++ b/hub/policy/src/test/java/uk/gov/ida/hub/policy/logging/HubEventLoggerTest.java
@@ -33,7 +33,6 @@ import java.util.UUID;
 
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.Mockito.verify;
-import static uk.gov.ida.eventemitter.EventDetailsKey.ab_test_variant;
 import static uk.gov.ida.eventemitter.EventDetailsKey.analytics_session_id;
 import static uk.gov.ida.eventemitter.EventDetailsKey.downstream_uri;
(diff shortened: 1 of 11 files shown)


JHjava:
@galund Yep whilst that is true for the analytic session_id we don't record the journey type on the response. Currently we need to go into the billing table to work out how many LOA2 user which have registered with Digi. After this Kibana should give us the same information and in theory is a lot easier to access and run queries against.

galund:
@JHjava it's really important to understand that this journey type is different to what you get if you look at the billing data. This just shows the user's intention - but the user can change their minds (i.e. journey switch from sign in to sign up, and vice versa).

This is explained a bit in an RFC, see the table under 'definition of success' where it shows 'what they said' vs 'what they did'. https://github.com/alphagov/verify-architecture/blob/master/rfcs/rfc-100-completion-rate.md

JHjava:
@galund Thanks for sharing that 👍 I'll have a read.

JHjava:
@galund OK that makes sense and the event system knows this based on the PID. I still think there's value to log it on the response.

PR #454, opened 2020-06-03
EID-1977 Validation of missing namespace uri on Assertion

idrop:
## What
Given an Assertion AttributeValue missing its QName namespace uri, use the matching namespace uri declared in the parent Response element to build a new QName containing a uri.

A match is considered when the namespace prefix of the Assertion and Response match e.g. they both have value of `eidas-natural`.

The new derived QName is then evaluated against a set of expected AttributeValue QNames.

The parent Response object need to be carried through a series of classes in order to be available on the `EidasAttributeStatementAssertionValidator`.

## How to review
Review `EidasAttributeStatementAssertionValidator` first, and debug the new unit test on `EidasAttributeStatementAssertionValidatorTest`, which contains an eidas Response in the desired state.

## Why
We have an issue validating the qualified name (QName) of decrypted AttributeValue received from the EU Validator Proxy Node, as they do not contain the namespace uri, which is required for evaluation against a set of expected QNames.

For example, given the AttributeValue below:
````
<saml2:Assertion>
......
    <saml2:AttributeValue xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
        xsi:type="eidas-natural:CurrentFamilyNameType">...
......
````
There is no definition for `eidas-natural` within the Assertion document.

The parent SAML Response XML document *does* contain the namespace uri i.e.
````
  <saml2p:Response .... 
      xmlns:eidas-natural="http://eidas.europa.eu/attributes/naturalperson" 
````
When evaluating assertions, we are only checking the `<Assertion>` SAML document, and have no reference to the namespace uri defined on the outer `<Response>` document.

### DIFF
--- a/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
@@ -3,6 +3,7 @@ package uk.gov.ida.hub.samlengine.validation.country;
 import org.opensaml.saml.saml2.core.Assertion;
 import org.opensaml.saml.saml2.core.Attribute;
 import org.opensaml.saml.saml2.core.AttributeStatement;
+import org.opensaml.saml.saml2.core.Response;
 import org.slf4j.event.Level;
 import uk.gov.ida.saml.core.IdaConstants.Eidas_Attributes;
 import uk.gov.ida.saml.core.errors.SamlTransformationErrorFactory;
@@ -15,10 +16,13 @@ import uk.gov.ida.saml.core.extensions.eidas.PersonIdentifier;
 import uk.gov.ida.saml.core.validation.SamlTransformationErrorException;
 import uk.gov.ida.saml.core.validation.SamlTransformationErrorManager;
 import uk.gov.ida.saml.core.validation.SamlValidationSpecificationFailure;
+import uk.gov.ida.saml.security.validators.ValidatedResponse;
 
+import javax.xml.XMLConstants;
 import javax.xml.namespace.QName;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -57,11 +61,11 @@ public class EidasAttributeStatementAssertionValidator {
     public EidasAttributeStatementAssertionValidator() {
     }
 
-    public void validate(Assertion assertion) {
-        validateAttributes(assertion);
+    public void validate(ValidatedResponse validatedResponse, Assertion assertion) {
+        validateAttributes(validatedResponse, assertion);
     }
 
-    private void validateAttributes(Assertion assertion) {
+    private void validateAttributes(ValidatedResponse validatedResponse, Assertion assertion) {
         final List<AttributeStatement> attributeStatements = assertion.getAttributeStatements();
         if (attributeStatements.isEmpty()) {
             SamlValidationSpecificationFailure failure = SamlTransformationErrorFactory.mdsStatementMissing();
@@ -95,9 +99,9 @@ public class EidasAttributeStatementAssertionValidator {
                 SamlValidationSpecificationFailure failure = SamlTransformationErrorFactory.emptyAttribute(attributeName);
                 throw new SamlTransformationErrorException(failure.getErrorMessage(), failure.getLogLevel());
             }
-            if (!VALID_TYPE_FOR_ATTRIBUTE.get(attributeName).equals(attribute.getAttributeValues().get(0).getSchemaType())) {
-                final QName schemaType = attribute.getAttributeValues().get(0).getSchemaType();
-                SamlValidationSpecificationFailure failure = SamlTransformationErrorFactory.attributeWithIncorrectType(attributeName, VALID_TYPE_FOR_ATTRIBUTE.get(attributeName), schemaType);
+            QName attributeValueSchemaType = getAttributeValueSchemaType(validatedResponse, attribute);
+            if (!VALID_TYPE_FOR_ATTRIBUTE.get(attributeName).equals(attributeValueSchemaType)) {
+                SamlValidationSpecificationFailure failure = SamlTransformationErrorFactory.attributeWithIncorrectType(attributeName, VALID_TYPE_FOR_ATTRIBUTE.get(attributeName), attributeValueSchemaType);
                 throw new SamlTransformationErrorException(failure.getErrorMessage(), failure.getLogLevel());
             }
             if (!VALID_ATTRIBUTE_NAME_FORMATS.contains(attribute.getNameFormat())) {
@@ -105,5 +109,20 @@ public class EidasAttributeStatementAssertionValidator {
                         invalidAttributeNameFormat(attribute.getNameFormat()));
             }
         }
+    }
+
+    private QName getAttributeValueSchemaType(ValidatedResponse response, Attribute attribute) {
+        final QName attributeValueSchemaType = attribute.getAttributeValues().get(0).getSchemaType();
+        if (XMLConstants.NULL_NS_URI.equals(attributeValueSchemaType.getNamespaceURI())) {
+            return response.getNamespaces().stream()
+                    .filter(ns -> Objects.equals(ns.getNamespacePrefix(), attributeValueSchemaType.getPrefix()))
+                    .findFirst()
+                    .map(ns -> new QName(
+                            ns.getNamespaceURI(),
+                            attributeValueSchemaType.getLocalPart(),
+                            attributeValueSchemaType.getPrefix()))
+                    .orElse(attributeValueSchemaType);
+        }
+        return attributeValueSchemaType;
     }
 }
--- a/hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java
+++ b/hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java
@@ -7,22 +7,30 @@ import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
+import org.opensaml.core.xml.Namespace;
 import org.opensaml.core.xml.XMLObject;
 import org.opensaml.saml.saml2.core.Assertion;
 import org.opensaml.saml.saml2.core.Attribute;
 import org.opensaml.saml.saml2.core.AttributeStatement;
+import org.opensaml.saml.saml2.core.AttributeValue;
+import uk.gov.ida.saml.core.IdaConstants;
 import uk.gov.ida.saml.core.IdaConstants.Eidas_Attributes;
 import uk.gov.ida.saml.core.extensions.eidas.CurrentFamilyName;
 import uk.gov.ida.saml.core.extensions.eidas.CurrentGivenName;
 import uk.gov.ida.saml.core.extensions.eidas.DateOfBirth;
 import uk.gov.ida.saml.core.extensions.eidas.PersonIdentifier;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java [hunk 1]
```diff
@@ -88,6 +90,26 @@ public void shouldThrowIfAttributeValueHasInvalidSchemaType() {
...
+    public void shouldThrowIfAttributeValueHasNullSchemaType() {
+        XMLObject xmlObject = mock(XMLObject.class);
+        when(firstName.getAttributeValues()).thenReturn(List.of(xmlObject));
+        when(xmlObject.getSchemaType()).thenReturn(null);
+
+        validator.validate(assertion);
+    }
+
+    @Test(expected = SamlTransformationErrorException.class)
+    public void shouldNotThrowIfAttributeValueHasNullSchemaNamespace() {
```
CharlesIC, line 103: `@Test(expected = ...)` indicates the test should throw an exception, but the name suggests it shouldn't (`shouldNotThrow...`). Which is it?

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java
```diff
@@ -88,6 +90,26 @@ public void shouldThrowIfAttributeValueHasInvalidSchemaType() {
...
+        XMLObject xmlObject = mock(XMLObject.class);
+        when(firstName.getAttributeValues()).thenReturn(List.of(xmlObject));
+        when(xmlObject.getSchemaType()).thenReturn(new QName(null, CurrentGivenName.TYPE_LOCAL_NAME, IdaConstants.EIDAS_NATURUAL_PREFIX));
+        when(firstName.getNameFormat()).thenReturn("urn:oasis:names:tc:SAML:2.0:attrname-format:uri");
```
CharlesIC, line 107: There's a SAML constant called `URI_REFERENCE` for `"urn:oasis:names:tc:SAML:2.0:attrname-format:uri"` in `org.opensaml.saml.saml2.core.Attribute` class

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
```diff
@@ -95,9 +95,11 @@ private void validateAttributes(Assertion assertion) {
                 SamlValidationSpecificationFailure failure = SamlTransformationErrorFactory.emptyAttribute(attributeName);
                 throw new SamlTransformationErrorException(failure.getErrorMessage(), failure.getLogLevel());
             }
-            if (!VALID_TYPE_FOR_ATTRIBUTE.get(attributeName).equals(attribute.getAttributeValues().get(0).getSchemaType())) {
-                final QName schemaType = attribute.getAttributeValues().get(0).getSchemaType();
-                SamlValidationSpecificationFailure failure = SamlTransformationErrorFactory.attributeWithIncorrectType(attributeName, VALID_TYPE_FOR_ATTRIBUTE.get(attributeName), schemaType);
+            QName expectedQName = VALID_TYPE_FOR_ATTRIBUTE.get(attributeName);
+            QName attributeValueQName = attribute.getAttributeValues().get(0).getSchemaType();
+            if (!isExpectedQName(expectedQName, attributeValueQName)) {
+                final QName schemaType = attributeValueQName;
```
CharlesIC, line 101:
```suggestion

```
This line is not needed - use `attributeValueQName` directly

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
```diff
@@ -106,4 +108,10 @@ private void validateAttributes(Assertion assertion) {
             }
         }
     }
+
+    private boolean isExpectedQName(QName expectedQName, QName attributeValueQName) {
+        return attributeValueQName != null
+                && expectedQName.getPrefix().equals(attributeValueQName.getPrefix())
+                && expectedQName.getLocalPart().equals(attributeValueQName.getLocalPart());
+    }
```
CharlesIC, line 116: This doesn't seem right to me. In this implementation we're not checking the attribute namespace, therefore, we're not checking that the attribute complies with the eIDAS schema.
I believe it's a standard XML practice to define namespaces at the top and then referring to them in children elements - so they can be referred to multiple times without re-defining each time. We should either:
- check if there's a ready XML/OpenSAML method to verify attribute namespaces
- create a map of namespaces in the in `Response` element and then verify the attribute namespace either exists in the attribute or in the map. E.g. for `xsi:type="eidas-natural:CurrentFamilyNameType"` the map should have an entry for key `eidas-natural`

I would check the XML/SAML/eIDAS spec to check where it's allowed to define namespaces to make sure we're looking in the right place.

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
```diff
@@ -106,4 +110,21 @@ private void validateAttributes(Assertion assertion) {
...
     }
+
+    private QName getAttributeValueSchemaType(Response response, Attribute attribute) {
+        final QName attributeValueSchemaType = attribute.getAttributeValues().get(0).getSchemaType();
+        if (attributeValueSchemaType.getNamespaceURI().isEmpty()) {
+            Optional<Namespace> matchingPrefix = response.getNamespaces()
+                    .stream()
+                    .filter(ns -> Objects.equals(ns.getNamespacePrefix(), attributeValueSchemaType.getPrefix()))
+                    .findFirst();
+            if (matchingPrefix.isPresent()) {
```
CharlesIC, line 121: ` .findFirst()` returns Optional so you can chain `ifPresent` directly on the result - no need for the second `if` statement

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java
```diff
@@ -95,7 +106,20 @@ public void shouldInsistOnMandatoryAttributes() {
...
-        validator.validate(assertion);
+        validator.validate(response, assertion);
+    }
+
+    @Test
+    public void validateEidasAttributesWhenEidasXMLNamespaceURIOnResponse() throws UnmarshallingException, XMLParserException {
+        Security.addProvider(new BouncyCastleProvider());
+        IdaSamlBootstrap.bootstrap();
+        InputStream responseIS = getClass().getClassLoader().getResourceAsStream("response-with-eidas-namespace.xml");
+        InputStream assertionIS = getClass().getClassLoader().getResourceAsStream("assertion-without-eidas-namespace.xml");
```
CharlesIC, line 117: We don't need XML resource files here - XML objects can be mocked. Other tests in this class do it so the approach can be copied in this test too

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
```diff
@@ -106,4 +110,21 @@ private void validateAttributes(Assertion assertion) {
             }
         }
     }
+
+    private QName getAttributeValueSchemaType(Response response, Attribute attribute) {
+        final QName attributeValueSchemaType = attribute.getAttributeValues().get(0).getSchemaType();
+        if (attributeValueSchemaType.getNamespaceURI().isEmpty()) {
+            Optional<Namespace> matchingPrefix = response.getNamespaces()
+                    .stream()
+                    .filter(ns -> Objects.equals(ns.getNamespacePrefix(), attributeValueSchemaType.getPrefix()))
```
CharlesIC, line 119: Why `Objects.equals` here? I'm suggesting this is the wrong method to use for comparison here, just curious why you've chosen it

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
idrop: I've now used Optional.map which is similar to ifPresent but allows a return

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
idrop: saves an extra NPE sanity check on `ns.getNamespacePrefix()`

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java
idrop: taken out now 👍  although I think using Mockito is perhaps more cognitive overhead than necessary here

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java
```diff
@@ -95,15 +108,46 @@ public void shouldInsistOnMandatoryAttributes() {
 
         when(attributeStatement.getAttributes()).thenReturn(List.of(firstName, lastName, dateOfBirth));
 
-        validator.validate(assertion);
+        validator.validate(response, assertion);
+    }
+
+    @Test
+    public void validateCorrectNamespaceURIOnResponseAllowsValidMatchOnAttribute() {
+
```
CharlesIC, line 116:
```suggestion
```
Seems like unnecessary whitespace?

hub/saml-engine/src/test/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidatorTest.java
```diff
@@ -95,15 +108,46 @@ public void shouldInsistOnMandatoryAttributes() {
...
+        validator.validate(response, assertion);
+
+        verify(response, times(1)).getNamespaces();
+
     }
 
+    @Test
+    public void validateIncorrectNamespaceURIOnResponseWhenNoneOnAttributeValueThrowsException() {
+
```
CharlesIC, line 131:
```suggestion
```

hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/validation/country/EidasAttributeStatementAssertionValidator.java
```diff
@@ -106,4 +110,19 @@ private void validateAttributes(Assertion assertion) {
...
+        final QName attributeValueSchemaType = attribute.getAttributeValues().get(0).getSchemaType();
+        if (XMLConstants.NULL_NS_URI.equals(attributeValueSchemaType.getNamespaceURI())) {
+            return response.getNamespaces().stream()
+                    .filter(ns -> Objects.equals(ns.getNamespacePrefix(), attributeValueSchemaType.getPrefix()))
+                    .findFirst()
+                    .map(ns -> new QName(
+                            ns.getNamespaceURI(),
+                            attributeValueSchemaType.getLocalPart(),
+                            attributeValueSchemaType.getPrefix()))
+                    .orElse(attributeValueSchemaType);
```
CharlesIC, line 124: Nice 👌

instantiator: This is magical. ✨

PR #455, opened 2020-06-10
HUB-601: Extend hub session duration

CharlesIC:
Extend hub session duration to reduce the proportion of people timing out. According to our research, extending the session to 2.5h from the current 90 mins will allow 20.8% of users timing out to complete their journey. We will measure this hypothesis and confirm the impact of the extension once this change has been deployed for a bit.

Details in [HUB-601](https://govukverify.atlassian.net/browse/HUB-601?focusedCommentId=25535&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25535)

Consider alongside alphagov/verify-frontend#872

### DIFF
--- a/hub/config/src/main/java/uk/gov/ida/hub/config/resources/CountriesResource.java
+++ b/hub/config/src/main/java/uk/gov/ida/hub/config/resources/CountriesResource.java
@@ -4,7 +4,6 @@ import com.codahale.metrics.annotation.Timed;
 import uk.gov.ida.hub.config.Urls;
 import uk.gov.ida.hub.config.data.LocalConfigRepository;
 import uk.gov.ida.hub.config.domain.CountryConfig;
-import uk.gov.ida.hub.config.exceptions.ExceptionFactory;
 
 import javax.inject.Inject;
 import javax.ws.rs.GET;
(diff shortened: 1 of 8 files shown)


hub/config/src/main/java/uk/gov/ida/hub/config/resources/CountriesResource.java
```diff
@@ -18,21 +17,16 @@
 public class CountriesResource {
 
     private final LocalConfigRepository<CountryConfig> countryConfigRepository;
-    private final ExceptionFactory exceptionFactory;
 
     @Inject
     public CountriesResource(
-            LocalConfigRepository<CountryConfig> countryConfigRepository,
-            ExceptionFactory exceptionFactory
```
galund, line 26: I think this change should be a separate commit

CharlesIC: `exceptionFactory` wasn't used in this class, hence removed. I really don't want to raise a separate PR for this given that it takes 2 days to get reviews

galund: Agree, these things don't need separate PR's, but separate commits means the intention is more clear to anyone looking at the history, and (sometimes) we have to back changes out, which makes separating stuff cleanly quite important, I think.

CharlesIC: ugh yeah separate commits is fair enough

galund:
ta :+1:

PR #456, opened 2020-06-11
📈 BAU Use a single counter to record connector metrics

idrop:
Use a single counter to record prometheus metrics on SAML requests and responses to EU Proxy Nodes, and classify the responses we receive.

The counter has the labels `country`, `direction` and `status`. We do not experience much traffic through the Connector Node, so the amount of data collected across all labels should be modest.

These metric will make it easier to create alerts and SLOs on questions like:

* is the OK response ratio acceptable per country
* is the request/response ratio acceptable per country
* are these ratios acceptable across countries

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/metrics/EidasConnectorMetrics.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/metrics/EidasConnectorMetrics.java
@@ -1,0 +1,47 @@
+package uk.gov.ida.hub.policy.metrics;
+
+import io.prometheus.client.Counter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URI;
+import java.util.Objects;
+
+public class EidasConnectorMetrics {
+
+    private static final Logger LOG = LoggerFactory.getLogger(EidasConnectorMetrics.class);
+
+    public enum Direction {request, response}
+
+    public enum Status {ok, ko, error}
+
+    private static final Counter CONNECTOR_COUNTER = Counter.build(
+            "verify_eidas_connector_count",
+            "EIDAS Connector Journey counter with labels 'country', 'direction', 'status'")
+            .labelNames("country", "direction", "status")
+            .register();
+
+    public static void increment(String entityId, Direction direction, Status status) {
+        try {
+            String country = getCountryCode(Objects.requireNonNull(entityId));
+            CONNECTOR_COUNTER.labels(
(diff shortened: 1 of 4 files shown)


hub/policy/src/main/java/uk/gov/ida/hub/policy/metrics/EidasConnectorMetrics.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static void increment(String entityId, Direction direction, Status status) {
+        try {
+            String country = getCountryCode(Objects.requireNonNull(entityId));
+            CONNECTOR_COUNTER.labels(
+                    country.toLowerCase(),
+                    Objects.requireNonNull(direction.name()),
+                    Objects.requireNonNull(status.name()))
+                    .inc();
+        } catch (Throwable e) {
```
galund, line 32: Is it standard to do this? If we think it's rare/not expected to ever happen, then it feels weird to turn it into a warning. Particularly _any_ throwable - doesn't Prometheus document what their library may throw?

idrop: The intent is to ensure that there's no way the `increment` method can break a user journey.
The Prometheus object is safe, the catching here is to deal with Objects.requireNonNull calls, and anything the `getCountryCode` method might throw.

The two exceptions that could be thrown by these methods are IllegalArgumentException & NullPointerException, so I've pushed a small change to narrow our expectation on the type of exception thrown

hub/policy/src/test/java/uk/gov/ida/hub/policy/metrics/EidasConnectorMetricsTest.java
```diff
@@ -0,0 +1,75 @@
...
+        when(counter.labels(expectedCountryCode, direction != null ? direction.name() : null, status != null ? status.name() : null)).thenReturn(counterChild);
+        EidasConnectorMetrics.increment(entityId, direction, status);
+        if (incrementsCounter) {
+            verify(counterChild).inc();
+        } else {
+            verifyNoInteractions(counter, counterChild);
+        }
+    }
+
+    private static void setFinalStatic(Field field, Object newValue) throws Exception {
```
galund, line 68: This is a bit gross :-) I found some possible alternative ways - there's a PowerMock thing perhaps - http://codyaray.com/2012/05/mocking-static-java-util-logger-with-easymocks-powermock-extension ?

PR #458, opened 2020-07-09
HUB-613: adjust timeouts between Microservices

jkunle:
See : https://govukverify.atlassian.net/browse/HUB-613

Redis client's timeout will need to change to something smaller than 30 seconds. The timeout among hub services would be 
Redis client = 20s, saml-engine=24 < policy = 26 < saml-proxy = 28 < frontend = 30secs

MSA is also a bottleneck with maximum timeout of about 13 seconds. It timeout is set at 24 seconds in this PR
also see https://github.com/alphagov/verify-frontend/compare/HUB-613-adjust-timeouts-between-microservices

### DIFF
--- a/configuration/policy.yml
+++ b/configuration/policy.yml
@@ -18,7 +18,7 @@ samlEngineUri: https://saml-engine.${DOMAIN}:443
 samlSoapProxyUri: https://saml-soap-proxy.${DOMAIN}:443
 enableRetryTimeOutConnections: true
 httpClient:
-  timeout: 60s
+  timeout: 26s
   timeToLive: 10m
   connectionTimeout: 4s
   keepAlive: 10s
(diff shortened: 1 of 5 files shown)


configuration/policy.yml
```diff
@@ -18,7 +18,7 @@ samlEngineUri: https://saml-engine.${DOMAIN}:443
 samlSoapProxyUri: https://saml-soap-proxy.${DOMAIN}:443
 enableRetryTimeOutConnections: true
 httpClient:
-  timeout: 60s
+  timeout: 26s
```
adityapahuja, line 21: There is SAML SOAP Proxy client's timeout in policy.yml that needs update. 
```
samlSoapProxyClient:
  timeout: 60s
```

PR #459, opened 2020-07-22
BAU: Update SAML Engine and Policy Redis's default timeout

adityapahuja:
According to the Lettuce (Redis Client) documentation, it says

```
 All connections inherit a default timeout from their {@link io.lettuce.core.RedisClient} and will throw a {@link io.lettuce.core.RedisException} when non-blocking commands fail to return a result before the timeout expires. The timeout defaults to 60 seconds and may be changed via {@link io.lettuce.core.RedisClient#setDefaultTimeout} or for each individual connection.
```

This change updates Redis client's default timeout from 60 seconds to 30 seconds so that SAML Engine and Policy can send exceptions back to the downstream applications when Redis client times out. Redis client's default timeout is configurable and is set to 20 seconds by default.

Author: @adityapahuja

### DIFF
--- a/hub/policy/src/main/java/uk/gov/ida/hub/policy/configuration/RedisConfiguration.java
+++ b/hub/policy/src/main/java/uk/gov/ida/hub/policy/configuration/RedisConfiguration.java
@@ -8,6 +8,7 @@ import java.net.URI;
 import java.time.Duration;
 
 import static java.time.temporal.ChronoUnit.MINUTES;
+import static java.time.temporal.ChronoUnit.SECONDS;
 
 public class RedisConfiguration {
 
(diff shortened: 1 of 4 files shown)


hub/policy/src/main/java/uk/gov/ida/hub/policy/configuration/RedisConfiguration.java
```diff
@@ -19,11 +20,19 @@
...
+    @Valid
+    @JsonProperty
+    private Duration timeout = Duration.of(20L, SECONDS);
+
     public Long getRecordTTL() {
         return recordTTL.getSeconds();
     }
 
     public RedisURI getUri() {
-        return RedisURI.Builder.redis(uri.getHost(), uri.getPort()).withTimeout(Duration.ofSeconds(20L)).build();
```
jkunle, line 27: `setDefaultTimeout(redisURI.getTimeout());` also gets called during the ` RedisClient redisClient = RedisClient.create()` but I guess leaving out the one for SAML Engine would have caused problems. I'll approve this

PR #462, opened 2020-07-24
Test Flakiness: Increase timeout to wait for a response to policy

idrop:
There is a [single integration test that is being flaky on big concourse with ~ 50% chance of passing](https://cd.gds-reliability.engineering/teams/verify/pipelines/deploy-verify-hub/jobs/build-verify-hub-microservices/builds/8).

Increase the timeout to wait for a response from stub MSA -> saml-soap-proxy -> policy, guessing that CPU/network are slower on this new concourse environment.


philandstuff:
having dug into the logs of the test failure, I think it's timing out trying to fetch metadata, not trying to talk to policy

idrop:
This has been fixed by amends to big concourse config instead

PR #471, opened 2020-09-09
EXErxpzy - Replace whitelist with allowlist

JHjava:
- We are replacing whitelist-repos with allowed-repos. We are running both for now so we can switch it over.

**To Review**
- Check that the URL for the allowed-repos can be accessed.


JohnWattsGDS (approved):
📃

Krenair (approved):

PR #472, opened 2020-09-11
HUB-675 Fixes for verify-local-startup

j-maynard:
In trying to get the docker stuff to work reliably I have modified the build.gradle file to have an integration test target as the verify-local-startup script requires it.  I have also added the empty file for the configuration of stub-event-sink.  This is also required by verify-local-startup.


instantiator (approved):

adityapahuja:
We don't need stub-event-sink. This application should be deprecated and removed from verify-hub.

j-maynard:
> We don't need stub-event-sink. This application should be deprecated and removed from verify-hub.

That's good to know and I am sure has been mentioned before... For this PR I am leaving it in place and will raise a ticket to tear it down and remove it from both here and the hub.

PR #475, opened 2020-10-01
BAU: Remove SHA1 logic

Wynndow:
**Not to be merged before this hub-fed-config PR: https://github.com/alphagov/verify-hub-federation-config/pull/575**

We no longer have any transactions using SHA1 in any environment. We can
remove the logic that dealt with it.


instantiator (approved):
That's a lot of hack and slash! ✨ magic

JohnWattsGDS (approved):
I don't know whether I should be happy or sad that we can delete 206 lines and only need four to make things right but have an approval anyway.

Wynndow:
@JohnWattsGDS Definitely happy. Definitely.

PR #477, opened 2020-10-05
BAU: Remove use of splunk logging library

Wynndow:
**Not to be merged until Cyber have confirmed their alerting is using the logs shipped via Cloudwatch rather than the logs sent directly by the Splunk appender**

We originally introduced the use of our own fork of Splunk's java
logging library when all the hubs logs were streamed to Kibana.

Now, all the hub's logs are sent Splunk instead. This means we can
remove the logic and the dependency on the library.


JohnWattsGDS (approved):
✂️

instantiator (approved):

PR #478, opened 2020-10-21
HUB-733 Add environment to sentry config

j-maynard:
This small config change should hopefully allow the hub apps to report their correct sentry environment to sentry.

### DIFF
--- a/configuration/config.yml
+++ b/configuration/config.yml
@@ -18,6 +18,7 @@ logging:
     - type: logstash-console
     - type: sentry
       dsn: ${SENTRY_DSN}
+      environment: ${SENTRY_ENV}
       threshold: ERROR
       tags: {"service-name": "config"}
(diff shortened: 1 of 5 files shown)


AmritSidhu (approved):

instantiator (approved):
Things I have learned this morning:

- the `logging.appender` of type `sentry` is picked up automatically by `dropwizard-sentry`
- we use `org.dhatim:dropwizard-sentry:1.3.9-2` (found in the main hub `build.gradle`)
- https://github.com/dhatim/dropwizard-sentry
- latest version depends on Dropwizard 2, so we're stuck on the 1.something version
- last updated 10 months ago (could be worse?)
- dhatim is an org, that forked this from dropwizard-raven
- their website, dhat.im is dead

I guess this is ok - but perhaps we need to include this library in a review of things we use some time?

PR #480, opened 2020-11-09
HUB-720 Fixes build issues with verify-local-startup

j-maynard:
This fixes the build issues with the stub-event-sink in verify-local-startup.  The stub-event-sink.yml is empty by design.  The way verify-local-startup works is it makes assumptions on how the individual hub components are built and the presence of their configuration files.  These changes bring stub-event-sink inline with the other hub components.


CharlesIC (commented):
I believe this new set of config should live in the `verify-local-startup` repo

Dockerfile
```diff
@@ -1,3 +1,5 @@
+ARG conf_dir=configuration
+ARG hub_app
```
philandstuff, line 2: I don't think we want `ARG` lines before `FROM`.  They are only visible in the `FROM` line itself, and the `FROM` doesn't reference them.

https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
```suggestion
```

Dockerfile
```diff
@@ -26,6 +28,7 @@ USER root
 ENV GRADLE_USER_HOME /usr/gradle/.gradle
 
 ARG hub_app
+ARG conf_dir
```
philandstuff, line 31: This ARG is not referenced anywhere, so we should remove it.

```suggestion
```

instantiator (approved):

Dockerfile
```diff
@@ -58,10 +61,11 @@ RUN gradle --console=plain \
 
 FROM openjdk:11.0.6-jre
 ARG hub_app
+ARG conf_dir
```
philandstuff, line 64:
```suggestion
ARG conf_dir=configuration
```
If we delete the lines at the top of the file, we need to reintroduce the default value here.

philandstuff (commented):
Looks sensible.  A few cleanup comments.

philandstuff (approved):

instantiator (approved):

PR #481, opened 2020-11-12
EID-2063: Config for EU exit date

AmritSidhu:
- Set `eidasDisabledAfter` as a LocalDateTime field on the hub config microservice
- Created a GET API endpoint at `/countries/eidasDisabledAfter` to receive this `LocalDateTime`
- Ensured that the `eidasDisabledAfter` config field can be overridden by env vars per environment

- Created AppRule tests to test the cases of, empty value, no config and null (empty is treated as null)
- Created a second `config.yml` for testing without the `eidasDisabledAfter` field called `config-no-eidas-exit-timestamp.yml`
- Within the tests the ports had to be specified (we used port 0 a wildcard) so that we could use multiple `configAppRules` with `configOverrides`

Co-authored-by: Phillip Miller <phillip.miller@digital.cabinet-office.gov.uk>

https://govukverify.atlassian.net/browse/EID-2063


hub/config/src/main/java/uk/gov/ida/hub/config/resources/CountriesResource.java
```diff
@@ -10,23 +11,35 @@
...
+        this.configuration = configuration;
     }
 
     @GET
     @Timed
     public Collection<CountryConfig> getEidasCountries() {
         return countryConfigRepository.getAllData();
     }
+
+    @Path(EIDAS_DISABLED_AFTER_ROOT)
```
idrop, line 39: Not sure this constant should be suffixed with `_ROOT`

hub/config/src/main/java/uk/gov/ida/hub/config/resources/CountriesResource.java
```diff
@@ -10,23 +11,35 @@
...
 @Produces(MediaType.APPLICATION_JSON)
 public class CountriesResource {
 
     private final LocalConfigRepository<CountryConfig> countryConfigRepository;
+    private final ConfigConfiguration configuration;
 
     @Inject
     public CountriesResource(
-            LocalConfigRepository<CountryConfig> countryConfigRepository) {
+            LocalConfigRepository<CountryConfig> countryConfigRepository, ConfigConfiguration configuration) {
```
idrop, line 28: I'd prefer to pass in the LocalDateTime on the constructor rather than the whole ConfigConfiguration. Be interested in @CharlesIC 's thoughts

idrop (approved):

instantiator (approved):
This is really clear, as are the tests.

PR #488, opened 2020-11-24
Revert "BAU Make REGISTRY_IMAGE_JDK a global build arg"

idrop:
Reverts alphagov/verify-hub#487

Received an error connecting to ECR from concourse task: `no basic auth credentials`


Wynndow:
Won't this revert to an also broken state?

jkunle (approved):

PR #489, opened 2020-11-27
HUB-774 Update the Dockerfile to use GHCR

j-maynard:
Update the Dockerfile to use Vendored containers in Github Container Registry to avoid issues with Docker Hub Rate limiting.


idrop (approved):
😄

jkunle (approved):
🥇

galund:
I'm not sure why you removed the ARGs as the DRY was a good thing... but obviously the principle of changing over to the new container registry is great :+1:

CharlesIC (approved):

CharlesIC:
@galund The args get reset at every `FROM` statement so you have to set them for every stage anyway. They weren't being used correctly before this change

PR #490, opened 2021-01-05
BAU - Rejecting unrecognised entityIds should come with a warning

instantiator:
Rejecting unrecognised `entityId`s is a correct behaviour. It should be logged as a warning, rather than an error.

This change alters logging behaviour for uncaught `NoKeyConfiguredForEntityException`s in `saml-proxy`, and when logging the creation of a `NotFoundException` when there's no data for a given `entityId` in `config`. In both cases, switching down from `ERROR` to `WARN` log entries.

FYI - I do not intend to merge this change until I've checked that cyber don't care. (ie. They _might_ consider unrecognised `entityId`s as some kind of indicator of malicious behaviour, and so they _might_ need to know that we're downgrading those logs from `ERROR` to `WARN`.)


j-maynard (approved):
🦁

JohnWattsGDS (approved):

PR #495, opened 2021-02-02
HUB-814: Send VSP version info to Prometheus

Wynndow:
We want to be able to easily see what version of the VSP services are
using.

This takes the version provided in authn requests and provides it to
Prometheus as a gauge.

The metric value of 1.0 is not the interesting part, as is only used as
a vehicle for the labels which describe the entity ID and the version.
It's only provided as metrics must provide a double value.

### DIFF
--- a/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
+++ b/hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
@@ -17,6 +17,7 @@ import io.lettuce.core.api.sync.RedisCommands;
 import io.lettuce.core.codec.RedisCodec;
 import io.lettuce.core.masterslave.MasterSlave;
 import io.lettuce.core.masterslave.StatefulRedisMasterSlaveConnection;
+import io.prometheus.client.Gauge;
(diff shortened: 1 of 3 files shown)


hub/saml-engine/src/main/java/uk/gov/ida/hub/samlengine/SamlEngineModule.java
```diff
@@ -172,6 +173,8 @@
     public static final String VERIFY_METADATA_RESOLVER = "VerifyMetadataResolver";
     public static final String FED_METADATA_ENTITY_SIGNATURE_VALIDATOR = "verifySignatureValidator";
     public static final String VERIFY_METADATA_SIGNATURE_TRUST_ENGINE = "VerifyMetadataSignatureTrustEngine";
+    public static final String VSP_VERSION_METRIC = "verify_saml_engine_vsp_version";
+    public static final String VSP_VERSION_METRIC_HELP = "VSP version reported from incoming Authn requests";
```
idrop, line 177:
```suggestion
    public static final String VSP_VERSION_METRIC_HELP = "VSP version by entityId, reported from incoming Authn requests";
```

idrop (approved):

instantiator (approved):

PR #503, opened 2021-03-08
EID-2106 Remove eIDAS connector node from Hub

idrop:
## Why

The UK is no longer part of the eIDAS regulation so we can remove the large eIDAS  connector node codebase to make the remaining Hub code and config easier to refactor and understand.

We no longer process eIDAS SAML in the frontend, most* eIDAS code and config will not currently be in use.

## What

* Remove all resources for processing eIDAS SAML requests and responses in the Hub.
* Remove all integration and unit tests for above

*Also ensure that the MSA Health Check to services continues to return the `eidasEnabled` flag, and we continue to export this boolean flag to Prometheus. This flag lets us know if the a service is sending the universal or legacy matching dataset to their local matching service.

## How

* this PR is perhaps not that intimidating 😏, for example 133 files are deleted in the first commit. Review in commit order.
* merge after the eIDAS frontend PR: https://github.com/alphagov/verify-frontend/pull/953

## Checks

✅ passed `pre-commit.sh`
✅ passed [acceptance tests](https://github.com/alphagov/verify-acceptance-tests) against a local hub using `SHOW_BROWSER=false TEST_ENV=local BROWSER=chrome ./pre-commit.sh`


hub/config/src/main/java/uk/gov/ida/hub/config/domain/TransactionConfig.java
```diff
@@ -65,14 +65,6 @@
     @JsonProperty
     protected List<LevelOfAssurance> levelsOfAssurance;
 
-    @Valid
-    @JsonProperty
```
Wynndow, line 69: There are transactions using these keys in their config files in verify-hub-federation-config. Is there any danger if things falling over if the key no longer exists here?

Wynndow (approved):
Everything being deleted is eIDAS related, and the tests are passing so I'm giving it a 👍 

I've not managed to get my local acceptance tests to run against this version of the hub though. Have you?

JohnWattsGDS (approved):
💀

hub/config/src/main/java/uk/gov/ida/hub/config/domain/TransactionConfig.java
idrop: When I start local hub using this `verify-hub` branch, and using the `ida-hub-acceptance-tests` [test-rp.yml transaction config](https://github.com/alphagov/ida-hub-acceptance-tests/blob/718130a4f950f193b95be2dfd6e33331115aa3c4/configuration/hub/config-service-data/transactions/test-rp.yml#L51) (on master)  it still has the `eidasEnabled=true` flag, and it starts ok.

I'm confident this transaction config file is used because [if I then set `enabled=false`](https://github.com/alphagov/ida-hub-acceptance-tests/blob/718130a4f950f193b95be2dfd6e33331115aa3c4/configuration/hub/config-service-data/transactions/test-rp.yml#L33), then the journey fails in the Hub.

PR #508, opened 2021-04-21
HUB-866: Add publish.gradle to Dockerfile

Wynndow:
We've added a new gradle file (`publish.gradle`) which is referenced in
the `build.gradle` file. If we don't copy the new file into the image
gradle breaks when being run during the build.


JohnWattsGDS (approved):
🗄️

idrop (approved):

PR #521, opened 2021-06-28
HUB-942: Remove use of dropwizard-guicey

Wynndow:
**It's probably best to review this commit by commit, but ignoring the first commit. It's a revert to bring in the previousuly approved DW2 work. The following two commits are the new work which fix the issues described below.**

This PR updates the previous PR (that has been merged and reverted a bunch of times), to remove the use of the dropwizard-guicey library.

Instead of using the library it shamelessly steals three classes from the DCS that do the same job: https://github.com/alphagov/doc-checking/tree/master/shared/src/main/java/uk/gov/ida/docchecking/shared/guice

This means we no longer need to import or worry about that 3rd party library.

This also has the effect of fixing an issue we noticed in staging and prod with the previous implementation. The symptoms of the issue were failures in validating responses from IDPs, due to the 'inResponseTo' attributes of assertions not matching the 'inResponseTo' of the response. This was caused by dropwizard-guicey effectively treating the validators as singletons rather than creating a new validator for every request. The resulted in a race condition with fields of the validator being overwritten by new requests before they could be read. The DCS implementation doesn't suffer from this and we use a new validator for each request.


idrop (approved):
I only reviewed the last two commits, and LGTM

JohnWattsGDS (approved):
🚀

PR #906, opened 2022-11-07
BAU - ITHC dockerfile fixes

dushan-madetech:
Only use the root user to assign permissions and ownership to the gradle user and use the gradle user for run commands etc.


elena-vi (approved):

owenblacker (approved):
LGTM

